PR #1, opened 2019-05-16
build: add ability to build only selected WASM runtimes.

PiotrSikora:
This commit adds 3 build-time options:

  1. bazel build -c opt --define wasm=enabled //source/...
  2. bazel build -c opt --define wasm=v8 //source/...
  3. bazel build -c opt --define wasm=wavm //source/...

which allow building either all or only selected WASM runtimes.

No WASM runtimes are built if the "wasm" option isn't provided.


source/extensions/common/wasm/BUILD
```diff
@@ -23,6 +25,7 @@ envoy_cc_library(
         ":well_known_names",
         "//include/envoy/http:codes_interface",
         "//include/envoy/http:filter_interface",
+        "//include/envoy/server:wasm_interface",
```
silentdai: wasm_interface is a dependency required even neither v8 nor wavm is selected. Not a big deal

source/extensions/common/wasm/wasm.cc
```diff
@@ -1763,11 +1767,18 @@ void GrpcStreamClientHandler::onRemoteClose(Grpc::Status::GrpcStatus status,
...
-  } else if (wasm_vm == WasmVmNames::get().Wavm) {
+  } else
+#endif
+#ifdef ENVOY_WASM_WAVM
+      if (wasm_vm == WasmVmNames::get().Wavm) {
     return Wavm::createVm();
-  } else {
+  } else
+#endif
+  {
```
silentdai: Tricky but works perfectly

test/extensions/filters/http/wasm/config_test.cc
```diff
@@ -33,7 +33,16 @@ class WasmFilterConfigTest : public TestBaseWithParam<std::string> {
   Api::ApiPtr api_;
 };
 
-INSTANTIATE_TEST_SUITE_P(Runtimes, WasmFilterConfigTest, testing::Values("wavm", "v8"));
+INSTANTIATE_TEST_SUITE_P(Runtimes, WasmFilterConfigTest,
+                         testing::Values(
+#if defined(ENVOY_WASM_V8) && defined(ENVOY_WASM_WAVM)
```
silentdai: This pattern repeats several times. Let's make it another macro on top of this.
Could be in another PR

source/extensions/common/wasm/BUILD
PiotrSikora: Yeah, that's because we're still building WASM filters with the abstraction layer, even without runtimes. I expect it to say this way, since we're getting null sandbox, which will be always compiled in.

PiotrSikora:
Updated, PTAL.

PR #5, opened 2019-05-21
wasm: provide working _emscripten_get_heap_size().

PiotrSikora:
While there, remove some code that assumed never-changing heap size.

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -486,7 +486,10 @@ void grpcSendHandler(void* raw_context, Word token, Word message_ptr, Word messa
   context->grpcSend(token, message, end_stream);
 }
 
-Word _emscripten_get_heap_sizeHandler(void*) { return std::numeric_limits<int32_t>::max(); }
+Word _emscripten_get_heap_sizeHandler(void* raw_context) {
+  auto context = WASM_CONTEXT(raw_context);
+  return context->wasmVm()->getMemorySize();
+}
 
 Word _emscripten_memcpy_bigHandler(void*, Word, Word, Word) {
   throw WasmException("emscripten emscripten_memcpy_big");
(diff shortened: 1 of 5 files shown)


source/extensions/common/wasm/wasm.cc
```diff
@@ -486,7 +486,10 @@ void grpcSendHandler(void* raw_context, Word token, Word message_ptr, Word messa
   context->grpcSend(token, message, end_stream);
 }
 
-Word _emscripten_get_heap_sizeHandler(void*) { return std::numeric_limits<int32_t>::max(); }
+Word _emscripten_get_heap_sizeHandler(void* raw_context) {
+  auto context = WASM_CONTEXT(raw_context);
```
jplevyak: The memory includes statics and the stack AFAICT.  Do you want to remove them?  That information is part of the emscripten metadata as well.  I have an old branch where I was doing the calculations for that, I'll see if I can dig it up.

PiotrSikora: I'm fairly certain that `emscripten_get_heap_size()` should return the size of the whole memory block, and the code in this PR seems to work fine (it doesn't break existing tests and fixes "memory access out of bounds" errors in V8 when loading `wasm-sd.wasm`).

For the reference:

Emscripten has it defined as:
```
emscripten_get_heap_size: function() {
  return HEAP8.length;
}
```

WAVM has it defined as:
```
DEFINE_INTRINSIC_FUNCTION(env, "_emscripten_get_heap_size", U32, _emscripten_get_heap_size)
{                                                                               
    Emscripten::Instance* instance = getEmscriptenInstance(contextRuntimeData);
    return coerce32bitAddress(instance->memory,                             
                              Runtime::getMemoryNumPages(instance->memory) * IR::numBytesPerPage);
}
```

PR #6, opened 2019-05-21
wasm: provide working writev() syscall.

PiotrSikora:
Implementation restricted to stdout/stderr redirected to Envoy logs.

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -535,7 +535,48 @@ Word ___syscall54Handler(void*, Word, Word) { throw WasmException("emscripten sy
 
 Word ___syscall140Handler(void*, Word, Word) { throw WasmException("emscripten syscall140"); }
 
-Word ___syscall146Handler(void*, Word, Word) { throw WasmException("emscripten syscall146"); }
+// Implementation of writev() syscall that redirects stdout/stderr to Envoy logs.
+// ssize_t writev(int fd, const struct iovec *iov, int iovcnt);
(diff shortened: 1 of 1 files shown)


PiotrSikora:
@jplevyak it doesn't appear that `__syscall4` is imported by any WASM modules, so I'd skip it for now.

jplevyak:
Yes, it looks like write() is the only function which calls SYS_write.

source/extensions/common/wasm/wasm.cc
```diff
@@ -535,7 +535,48 @@ Word ___syscall54Handler(void*, Word, Word) { throw WasmException("emscripten sy
...
+    const uint32_t* iovec = reinterpret_cast<const uint32_t*>(memslice.data());
+    if (iovec[1] /* size */) {
+      memslice = context->wasmVm()->getMemory(iovec[0] /* data */, iovec[1] /* size */);
+      s.append(memslice.data(), memslice.size());
+    }
+  }
+
+  size_t written = s.size();
+  if (written) {
+    // Remove trailing newline from the logs, if any.
```
silentdai: I would be surprised when I get the same log despite I write with "\n" or without "\n". TBH I don't think it is the syscall layer to handle this logic.

PiotrSikora: We're intercepting writes to stdout/stderr and writing them to Envoy logs, where each entry already has a newline at the end. If we don't remove those trailing newlines, then each intercepted write to stdout/stderr will be followed by an empty line.

I don't think that removing empty lines is more intrusive than prefixing those writes with a timestamp, filename and `wasm log:`, so I'm going to leave this as-is.

PR #8, opened 2019-05-21
v8: log calls to host functions.

PiotrSikora:

### DIFF
--- a/source/extensions/common/wasm/v8/v8.cc
+++ b/source/extensions/common/wasm/v8/v8.cc
@@ -37,6 +37,16 @@ wasm::Engine* engine() {
   return engine.get();
 }
 
+struct FuncData {
+  FuncData(std::string name) : name(name) {}
+
+  std::string name;
+  wasm::own<wasm::Func*> callback;
+  void* raw_func;
+};
+
+typedef std::unique_ptr<FuncData> FuncDataPtr;
+
 class V8 : public WasmVm {
 public:
   V8() = default;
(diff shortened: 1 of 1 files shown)


source/extensions/common/wasm/v8/v8.cc
```diff
@@ -552,39 +562,49 @@ template <typename... Args>
...
         auto args_tuple = convertValTypesToArgsTuple<std::tuple<Args...>>(params);
         auto args = std::tuple_cat(std::make_tuple(current_context_), args_tuple);
-        auto function = reinterpret_cast<void (*)(void*, Args...)>(data);
+        auto function = reinterpret_cast<void (*)(void*, Args...)>(func_data->raw_func);
         absl::apply(function, args);
         return nullptr;
       },
-      reinterpret_cast<void*>(function));
-  host_functions_.emplace(absl::StrCat(moduleName, ".", functionName), std::move(func));
+      data.get());
```
jplevyak: You don't want to use a more sophisticated constructor and emplace?  Trying to avoid piecewise constructors?

jplevyak: You could also use default constructor and a copy constructor.

PiotrSikora: Which object are you talking about? `FuncData`? It's kind of ugly, because it needs to include pointer to `callback` created using `wasm::Func::make()` which includes pointer to said `FuncData`. I could set `raw_func` in constructor, but I left it here for readability.

I'm fine changing things if you can provide code snippets, but I'm not sure what changes do you want to see, exactly.

jplevyak: Ah, I didn't see the recursive reference in the lambda.  LGTM

PR #12, opened 2019-05-24
Add grpc service into intrinsic full proto

bianpengyuan:


bianpengyuan:
cc @PiotrSikora @jplevyak

silentdai:
Why are the generated pb.h and pb.cc tracked by git?

bianpengyuan:
Because we want to pin the version of generated code (3.6.1) so that we can link with prebuilt protobuf library. See readme here: https://github.com/envoyproxy/envoy-wasm/blob/master/api/wasm/cpp/README.md

bianpengyuan:
Created #17 to track this.

api/wasm/cpp/proxy_wasm_intrinsics.proto
```diff
@@ -1,7 +1,82 @@
...
+      DataSource cert_chain = 3;
+    }
+    message GoogleLocalCredentials {
+    }
+    message Empty {
+    }
+    message ChannelCredentials {
+      oneof credential_specifier {
+        SslCredentials ssl_credentials = 1;
+        Empty google_default = 2;
```
PiotrSikora: Why can't we use `google.protobuf.Empty google_default = 2;` here? Seems to work fine for me.

api/wasm/cpp/proxy_wasm_intrinsics.proto
```diff
@@ -1,7 +1,82 @@
...
+    repeated CallCredentials call_credentials = 3;
+    string stat_prefix = 4;
+    string credentials_factory_name = 5;
+    google.protobuf.Struct config = 6;
+  }
+  oneof target_specifier {
+    EnvoyGrpc envoy_grpc = 1;
+    GoogleGrpc google_grpc = 2;
+  }
+  reserved 4;
```
PiotrSikora: `google.protobuf.Duration timeout = 3;` is missing, compared to upstream.

api/wasm/cpp/proxy_wasm_intrinsics.pb.h
```diff
@@ -1,8 +1,8 @@
 // Generated by the protocol buffer compiler.  DO NOT EDIT!
-// source: proxy_wasm_intrinsics.proto
+// source: api/wasm/cpp/proxy_wasm_intrinsics.proto
```
PiotrSikora: How did you generate this? Somehow, the path and object names changed to include the directory, compared to the version that's currently checked in (though, I'm not sure if this matters at all).

api/wasm/cpp/proxy_wasm_intrinsics.proto
bianpengyuan: Done.

api/wasm/cpp/proxy_wasm_intrinsics.proto
bianpengyuan: Added. @jplevyak The lite proto does not have this, it is to avoid having a copy of Duration proto?

api/wasm/cpp/proxy_wasm_intrinsics.pb.h
bianpengyuan: I ran `protoc --cpp_out=./ -I./ -Iapi/wasm/cpp  api/wasm/cpp/proxy_wasm_intrinsics.proto` under root directory to generate the code. I think it does not matter, when building with null, the plugin uses GrpcService from upstream. The generated code is used in wasm module, where include path depends. But I think it is better to keep it in sync with lite version, so I reran the command within `api/wasm/cpp` and now those path diff should be gone.

PiotrSikora:
Should we merge this, @bianpengyuan? cc @jplevyak

bianpengyuan:
Yes, it would be good to get this in for consistency between lite and full versions.

PR #25, opened 2019-05-30
[WIP] metadata_exchange: pass node metadata between two proxies

kyessenov:
Working version of a basic metadata exchange filter that populates downstream and upstream metadata using request/response headers.

Example log:
```
server 200 downstream:{"app":"productpage"} upstream:-
client 200 downstream:- upstream:{"app":"productpage"}
```

/assign @mandarjog 
/assign @douglas-reid


bianpengyuan:
Seems like now metadata is stored into host. It would be nice if we could have it stored in VM's memory as well so that downstream filters that share the same VM does not need to call out to host and copy/cache the data again. John is going to add mechanism for communication between wasm instances, so it seems feasible to me. Do you already plan on doing that?

kyessenov:
@bianpengyuan What's the relationship between VM memory and module memory? If we want security properties, then one module should not access another module directly without host involvement. So in the final architecture of sandboxed modules, it seems reasonable to go through the host and delegate ownership of the request-bound data to it. I don't have a clear view of how VMs, host, contexts, and modules all fit together, so it would be good to write-it up somewhere.

kyessenov:
fyi @gargnupur

douglas-reid:
@kyessenov do you have brief one-pager writeup on how the peer metadata exchange is meant to work (and to be tested)?

bianpengyuan:
@kyessenov I am not sure about that. @jplevyak @PiotrSikora For modules running on the same VM, is their memory space isolated? If not, do we want to have an API to support VM internal data sharing so that we can avoid an extra copy of metadatas for each request?

The current spec maps vm 1:1 to a wasm module, so there seems to be no good way to share a VM between modules with different code (e.g. metadata exchange and stackdriver), unless squashing their code by making this metadata exchange a library. I could write something up about the wasm runtime and lifecycle of objects. I lost track every time after reading the code for a while so I think it would be good to have a doc about that.

kyessenov:
@bianpengyuan I think that is expected since sandboxing carries a cost for passing data safely between modules. We _should_ minimize processing costs per transfer. From my code inspection, there is always protobuf serialization hidden anytime you access metadata, and that code should not be hidden.

kyessenov:
@douglas-reid yes, it's coming. I wanted an early investigation since we're building on top of unstable APIs without design docs for them :)

Issue #18, opened 2019-05-29
v8 does not build on macos


bianpengyuan:
v8 cannot be built on mac due to a link error, which makes istio-proxy CI not green: https://github.com/istio/proxy/pull/2231.

I can get the same output as circleci on my mac locally:

```
ld: warning: could not create compact unwind for _forkedStackTrampoline: does not use RBP or RSP based frame
Undefined symbols for architecture x86_64:
  "wasm::vec<wasm::ExportType*>::free_data()", referenced from:
      wasm::vec<wasm::ExportType*>::~vec() in libv8_lib.lo(v8.o)
  "wasm::vec<wasm::ImportType*>::free_data()", referenced from:
      wasm::vec<wasm::ImportType*>::~vec() in libv8_lib.lo(v8.o)
  "wasm::vec<wasm::Extern*>::free_data()", referenced from:
      wasm::vec<wasm::Extern*>::~vec() in libv8_lib.lo(v8.o)
  "wasm::vec<wasm::ValType*>::free_data()", referenced from:
      wasm::vec<wasm::ValType*>::~vec() in libv8_lib.lo(v8.o)
  "wasm::vec<wasm::ValType*>::make_data()", referenced from:
      wasm::vec<wasm::ValType*>::vec(unsigned long) in libv8_lib.lo(v8.o)
  "wasm::vec<char>::free_data()", referenced from:
      wasm::vec<char>::~vec() in libv8_lib.lo(v8.o)
      wasm::vec<char>::reset() in libv8_lib.lo(v8.o)
  "wasm::vec<char>::make_data()", referenced from:
      wasm::vec<char>::vec(unsigned long) in libv8_lib.lo(v8.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

cc @PiotrSikora

bianpengyuan:
Looks like the recipe only builds with x64 instead of ia32 (https://github.com/envoyproxy/envoy-wasm/blob/master/ci/build_container/build_recipes/v8.sh#L42). I will try to update the recipe to also supports mac.

PiotrSikora:
@bianpengyuan but macOS is x86_64 nowadays.

Also, it definitely did built for macOS correctly at some point using this recipe, so something broke... I'll try to dig into this later today, but this might take a while due to build times :(

PiotrSikora:
[...]

PR #26, opened 2019-05-30
build: fix linking against wasm-c-api on macOS.

PiotrSikora:
While there, re-enable make/free accounting, since it was broken
for the same reason (implementation was leaked into headers).

Fixes envoyproxy/envoy-wasm#18.

### DIFF
--- a/ci/build_container/build_recipes/v8.sh
+++ b/ci/build_container/build_recipes/v8.sh
@@ -72,23 +72,7 @@ cat <<\EOF | patch -p1
 
 EOF
 
-# 2. Force full GC when destroying VMs.
-
-cat <<\EOF | patch -p1
---- a/src/wasm-v8.cc
-+++ b/src/wasm-v8.cc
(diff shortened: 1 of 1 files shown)


PiotrSikora:
cc @bianpengyuan

ci/build_container/build_recipes/v8.sh
```diff
@@ -53,29 +53,26 @@ cd ../..
...
  CC_FLAGS = ${C_FLAGS}
 -LD_FLAGS = -fsanitize-memory-track-origins -fsanitize-memory-use-after-dtor
 +LD_FLAGS =
 
  C_COMP = clang
+
 EOF
 
-# 3. Force full GC when destroying VMs.
+# 2. Force full GC when destroying VMs.
```
jplevyak: Now that DEBUG is set can you revert the change to #if 1 ?

PiotrSikora: Good catch, thanks! Done.

PR #28, opened 2019-05-31
Expose Node attributes. Support Listener metadata.  Support Cluster M…

jplevyak:
…etadata.

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -1017,25 +1017,22 @@ const ProtobufWkt::Struct* Context::getMetadataStructProto(MetadataType type,
     }
     return nullptr;
   case MetadataType::Cluster: {
+    std::string cluster_name;
     auto stream_info = getConstStreamInfo(type);
     if (!stream_info) {
       return nullptr;
     }
-    auto router_entry = stream_info->routeEntry();
-    if (!router_entry) {
+    auto host = stream_info->upstreamHost();
+    if (!host) {
       return nullptr;
     }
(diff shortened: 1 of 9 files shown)


jplevyak:
This only supports the "local cluster".  I presume we would like to support arbitrary cluster metadata by name.  I am also looking for feedback on the non-metadata metadata (e.g. things in Node) which I am exposing here with the special domain "-".  Again, could do this a different way, looking for feedback, ideas.

source/extensions/common/wasm/wasm.cc
```diff
@@ -983,7 +983,42 @@ const ProtobufWkt::Struct* Context::getMetadataStructProto(MetadataType type,
   case MetadataType::ResponseRoute:
     return getRouteMetadataStructProto(encoder_callbacks_);
   case MetadataType::Node:
+    if (name == "_") {
```
kyessenov: I think it would be better not to supply all the metadata at once, but just have some kind of switch based on keys that only returns individual fields.

jplevyak: You can ask for individual bits of the metadata with the high level API by calling namedMetadataValue("_", "cluster") for example.

jplevyak:
I updated this with two different ways to get the upstream cluster name.  Please take a look.

I also changed the way to get "builtin" metadata (e.g. Node attributes) to use the name ".".

jplevyak:
Updated wrt latest discussion about how to get the upstream cluster.  PTAL

PR #29, opened 2019-05-31
wasm: use values extracted from Emscripten metadata.

PiotrSikora:
This allows us to use more recent versions of Emscripten.

Fixes envoyproxy/envoy-wasm#32.

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -499,11 +499,11 @@ Word _emscripten_resize_heapHandler(void*, Word) {
   throw WasmException("emscripten emscripten_resize_heap");
 }
 
-Word abortOnCannotGrowMemoryabi0_0Handler(void*) {
+Word abortOnCannotGrowMemoryAbi00Handler(void*) {
   throw WasmException("emscripten abortOnCannotGrowMemory");
 }
 
-Word abortOnCannotGrowMemoryabi0_2Handler(void*, Word) {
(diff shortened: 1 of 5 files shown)


source/extensions/common/wasm/wasm.cc
```diff
@@ -1324,16 +1329,24 @@ Wasm::Wasm(absl::string_view vm, absl::string_view id, absl::string_view initial
...
-      "envoy", #_fn, &_fn##Handler,                                                                \
-      &ConvertFunctionWordToUint32<decltype(_fn##Handler),                                         \
-                                   _fn##Handler>::convertFunctionWordToUint32);
+      "envoy", #_fn, &_fn##_abi##Handler,                                                          \
+      &ConvertFunctionWordToUint32<decltype(_fn##_abi##Handler),                                   \
+                                   _fn##_abi##Handler>::convertFunctionWordToUint32)
+#define _REGISTER(_fn) _REGISTER_ABI(_fn, )
+
   if (is_emscripten_) {
+    if (emscripten_abi_major_version_ > 0 || emscripten_abi_minor_version_ > 1) {
```
jplevyak: Do you want to check the precise version and then fail to load the module if it isn't one of the know versions?

PiotrSikora: I don't, because then we'd stop loading working WASM modules (e.g. neither change to ABI 0.1 nor 0.3 required any changes in Envoy-WASM)... unless you feel strongly about it?

jplevyak: I'm OK with it this way.

PR #33, opened 2019-06-04
wasm: import functions from the module before starting VM.

PiotrSikora:
This prevents a nullptr-dereference when global constructors
call host function which calls module function (e.g. malloc).


jplevyak:
Why would a global constructor call a host function ? (just for my edification).

PiotrSikora:
Because they need to interact with the host environment.

e.g. `wasm-sd.wasm`'s global constructors call `emscripten_get_heap_size()`, `pthread_equal()`, and `cxa_allocate_exception()` (when throwing an exception), which in turn might call module's `malloc()`.

PR #35, opened 2019-06-04
Switch onDelete and onDone call

bianpengyuan:
It feels to me that onDone() should be called at the end of log callback which means on more callback is going to be triggered, and onDelete should be called in onDesdroy when filter is deleted. This would make code fit better with the semantic.

@jplevyak @PiotrSikora Feel free to close this PR if the original way is preferred. Thanks!

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -1558,14 +1558,14 @@ void Context::log(const Http::HeaderMap* request_headers, const Http::HeaderMap*
   access_log_response_headers_ = nullptr;
   access_log_stream_info_ = nullptr;
 
-  onDelete();
+  onDone();
 }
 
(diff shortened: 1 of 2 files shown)


jplevyak:
So, onDone means that the operations has completed.  This is when the stream proper has closed and can no longer change state.  After that the onLog call is made which is used for logging the results of the completed operation.   Finally the onDelete calls is called.  This is when no further calls can come in.  It is the opportunity to free up any state (e.g. the Context).  I think it is better the way that it is, but that we clearly need some better documentation.    The onDelete can't be called in onDestroy because that is just for the HttpFilter, and we can still get onLog calls after that.

bianpengyuan:
@jplevyak I see. I thought onDestroy happens after log inside envoy, which made the ordering a bit weird. Thanks for clarification!

PR #38, opened 2019-06-07
build: update v8 to 7.5.288.22 and migrate to wee8.

PiotrSikora:

### DIFF
--- a/ci/build_container/build_recipes/v8.sh
+++ b/ci/build_container/build_recipes/v8.sh
@@ -2,21 +2,6 @@
 
 set -e
 
-# Get wasm-c-api.
-
-COMMIT=dc8cc29340e0c374dd350e0de8ff4b22d429af1e  # Thu May 16 21:03:37 2019 +0200
-SHA256=dea56d64f58110aee585857bca279701d39bdcb2dc70bdb8eb02fe6a8571c714
-
-curl https://github.com/WebAssembly/wasm-c-api/archive/"$COMMIT".tar.gz -sLo wasm-c-api-"$COMMIT".tar.gz \
-  && echo "$SHA256" wasm-c-api-"$COMMIT".tar.gz | sha256sum --check
-tar xf wasm-c-api-"$COMMIT".tar.gz
-cd wasm-c-api-"$COMMIT"
-
-# Build v8 inside v8 subdirectory to match wasm-c-api's Makefile.
-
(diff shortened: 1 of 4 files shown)


ci/build_container/build_recipes/v8.sh
```diff
@@ -2,133 +2,55 @@
...
- DEFINE_VEC(Extern*, EXTERN)
- DEFINE_VEC(Val, VAL)
-
--#endif  // #ifdef DEBUG
--
-
- ///////////////////////////////////////////////////////////////////////////////
- // Runtime Environment
-
++#if 1
```
jplevyak: Remove ?

ci/build_container/build_recipes/v8.sh
```diff
@@ -2,133 +2,55 @@
...
+     reinterpret_cast<i::Isolate*>(isolate_)->heap()->PreciseCollectAllGarbage(
+         i::Heap::kNoGCFlags, i::GarbageCollectionReason::kTesting,
+         v8::kGCCallbackFlagForced);
 EOF
 
-# Build wasm-c-api.
-
-# TODO(PiotrSikora): respect CC/CXX/CFLAGS/CXXFLAGS/LDFLAGS upstream.
+# Build wee8 library.
```
jplevyak: Could you comment on what is wee8 vis-a-vis v8 ?

ci/build_container/build_recipes/v8.sh
PiotrSikora: We need this (again), since we're not building whole v8/wee8 with `DEBUG`, so it's leaking memory by default.

ci/build_container/build_recipes/v8.sh
PiotrSikora: wee8 is supposed to be a trimmed down version of v8 that supports only WebAssembly (i.e. no JavaScript), but since this initiative just started, right now wee8 == v8.

ci/build_container/build_recipes/v8.sh
jplevyak: I was wondering if you wanted to keep the "#if 1" or remove it?

PiotrSikora: Ah, let's leave it, otherwise I'd need to also remove `#endif`, which leads to a larger diff.

PR #40, opened 2019-06-07
Fix duplicated assignment to response header in log

bianpengyuan:


bianpengyuan:
cc @jplevyak @PiotrSikora

PR #41, opened 2019-06-10
Add basic doc for wasm API

bianpengyuan:
Add some basic doc as reference for WASM API. Still bunch of todos such as metadata, vm sharing and out of tree example, high level design thinking etc.

@jplevyak @PiotrSikora @mandarjog I am not sure what is the good way to make this more reviewable since there is not CI. You may need to build the doc locally to get a better look.


docs/root/configuration/http_filters/wasm_filter.rst
```diff
@@ -0,0 +1,1184 @@
+.. _config_http_filters_wasm:
```
PiotrSikora: This documentation should really be part of the C++ SDK, not Envoy proper.

bianpengyuan: Where is c++ sdk? First time heard about his.

PiotrSikora: `api/wasm/cpp`

docs/root/configuration/http_filters/wasm_filter.rst
```diff
@@ -0,0 +1,1184 @@
...
+
+Makes an gRPC stream to an upstream host.
+
+*service* is a serialized proto string of :ref:`gRPC service <envoy_api_msg_core.GrpcService>` for gRPC client initialization.
+*service_name* and *method_name* indicates the target gRPC service and method name.
+*handler* (:ref:`GrpcStreamHandler <config_http_filters_wasm_GrpcStreamHandler>`) is used to control the stream and
+as target for gRPC stream callbacks.
+
+.. note::
+    if the stream call outlives the per request context, *grpcStreamHandler* should be called within :ref:`root context <config_http_filters_wasm_context_object>`.
```
jplevyak: This is an issue.  There is no easy way to create a callback on another context_id.  Moreover onDelete will delete the Context making it impossible to get the gRPC result.   We could decouple the gRPC upcalls from the stream context and always invoke them with the root context id.  The high level code would then invoke a std::function which could do whatever, including attempt to find a context with (a new) getContext(stored_context_id) call.   Alternatively I can block the doDelete call until the last http/gRPC callback has arrived... thoughts?

docs/root/configuration/http_filters/wasm_filter.rst
```diff
@@ -0,0 +1,1184 @@
...
+setTickPeriodMilliseconds
+^^^^^^^^^^^^^^^^^^^^^^^^^
+.. code-block:: cpp
+
+    void setTickPeriodMilliseconds(uint32_t millisecond)
+
+Set a timer. *millisecond* is tick interval in millisecond. :ref:`onTick <config_http_filters_wasm_context_object_api_ontick>` will be invoked when timer fires.
+
+.. note::
+    Only one timer could be set for each WASM module, so it needs to be multiplexed by events with different tick intervals.
```
jplevyak: Post https://github.com/envoyproxy/envoy-wasm/pull/44 one timer per root_id which is vectored to the appropriate RootContext.   Any context can call setTickPeriodMilliseconds and the onTick will come on the corresponding RootContext.

docs/root/configuration/http_filters/wasm_filter.rst
```diff
@@ -0,0 +1,1184 @@
...
+to determine how filter chain iteration proceeds.
+
+onDone
+^^^^^^
+
+.. code-block:: cpp
+
+    void onDone()
+
+Called after stream is ended or reset. All stream info will not be changed any more and is safe for access logging.
```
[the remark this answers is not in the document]
mandarjog: When should you choose onDone vs onLog?

docs/root/configuration/http_filters/wasm_filter.rst
```diff
@@ -0,0 +1,1184 @@
...
+
+Makes an HTTP call to an upstream host. 
+
+*cluster* is a string which maps to a configured cluster manager cluster. 
+*request_headers* is a vector of key/value pairs to send. Note that the *:method*, *:path*, and *:authority* headers must be set. 
+*request_body* is an optional string of body data to send. timeout is an integer that specifies the call timeout in milliseconds. 
+*timeout_milliseconds* is an unsigned integer as timeout period for the http call in milliseconds.
+*callback* is the callback function to be called when the HTTP request finishes.
+
+.. note::
```
mandarjog: Can we enforce this thru API? Otherwise a user will see sigsegv randomly.

jplevyak:
Let's see if we can get this in even if it isn't perfect so that we can start editing it incrementally via smaller reviews.    Piotr where should this be put?

bianpengyuan:
@jplevyak @PiotrSikora  I am actually working on moving this doc into SDK directory. Will update it today.

docs/root/configuration/http_filters/wasm_filter.rst
bianpengyuan: I created a doc/ folder under api/wasm/cpp and move this doc to there.

docs/root/configuration/http_filters/wasm_filter.rst
```diff
@@ -0,0 +1,1184 @@
...
+* This filter should be configured with name *envoy.wasm*.
+
+Example
+-------
+
+An example C++ WASM filter could be found `here <https://github.com/envoyproxy/envoy-wasm/tree/19b9fd9a22e27fcadf61a06bf6aac03b735418e6/examples/wasm>`_.
+
+To implement a WASM filter:
+
+* Create a `filter context class <https://github.com/envoyproxy/envoy-wasm/blob/19b9fd9a22e27fcadf61a06bf6aac03b735418e6/examples/wasm/envoy_filter_http_wasm_example.cc#L7>`_ which inherits the `base context class <https://github.com/envoyproxy/envoy-wasm/blob/19b9fd9a22e27fcadf61a06bf6aac03b735418e6/api/wasm/cpp/proxy_wasm_impl.h#L225>`_.
```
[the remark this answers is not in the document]
bianpengyuan: I added root context into the steps

docs/root/configuration/http_filters/wasm_filter.rst
bianpengyuan: I think it makes sense to hide this in API and make calls from stream context always attach to root context.

bianpengyuan: For this PR though I think we could leave it as what it is and remove the notes when the improvement is made? I can create a issue for this.

docs/root/configuration/http_filters/wasm_filter.rst
bianpengyuan: See this comment thread: https://github.com/envoyproxy/envoy-wasm/pull/41#discussion_r293609704 it is of the same topic.

docs/root/configuration/http_filters/wasm_filter.rst
```diff
@@ -0,0 +1,1184 @@
...
+
+.. _config_http_filters_wasm_context_object:
+
+Context object
+--------------
+
+WASM module is running in a stack-based virtual machine and its memory is isolated from the host environment. 
+All interactions between host and WASM module are through functions and callbacks wrapped by context object. 
+
+At bootstrap time, a root context with id 0 is created.
```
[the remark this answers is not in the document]
bianpengyuan: I would still prefer it to be root context. vm context is not exposed to user, add a vm context here seems to introduce some confusion.

docs/root/configuration/http_filters/wasm_filter.rst
bianpengyuan: https://github.com/envoyproxy/envoy-wasm/issues/85 Issue created.

docs/root/configuration/http_filters/wasm_filter.rst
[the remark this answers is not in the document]
jplevyak: Right.  Now context 0 is only used for non-API calls into the VM (e.g. malloc so we can setup an API call).

So this should be "At bootstrap time, a root context is created.".

There is now a root context per root_id in the configuration (typically only one root context for simple cases).  We can discuss the more complicated case in a dedicated section.

docs/root/configuration/http_filters/wasm_filter.rst
jplevyak: LGTM

docs/root/configuration/http_filters/wasm_filter.rst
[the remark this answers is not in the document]
bianpengyuan: Sg, updated.

jplevyak:
I think we should get this in and then do smaller PRs to clean it up.  I don't want to saddle Pengyuan with doing all the edits.

PR #47, opened 2019-06-11
Wasm queues

jplevyak:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


jplevyak:
This is an early peek of the queues API.  I am going to complete the tests and update this PR.

This is based on the root context PR, so the commit you want to look at is:

https://github.com/envoyproxy/envoy-wasm/pull/47/commits/560c118ab3b2a46024c2d61bf44bd02e070c1afd

jplevyak:
This PR is complete and ready for review

api/wasm/cpp/proxy_wasm_externs.h
```diff
@@ -16,6 +16,7 @@
    extern "C" EMSCRIPTEN_KEEPALIVE void proxy_onStart(uint32_t root_context_id, uint32_t root_id_ptr, uint32_t root_id_size);
    extern "C" EMSCRIPTEN_KEEPALIVE void proxy_onConfigure(uint32_t root_context_id, uint32_t configuration_ptr, uint32_t configuration_size);
    extern "C" EMSCRIPTEN_KEEPALIVE void proxy_onTick(uint32_t root_context_id);
+   extern "C" EMSCRIPTEN_KEEPALIVE void proxy_onQueueReady(uint32_t root_context_id, uint32_t queue_name_ptr, uint32_t queue_name_size);
```
PiotrSikora: Nit: `proxy_onSharedQueueReady`.

api/wasm/cpp/proxy_wasm_externs.h
```diff
@@ -85,6 +86,18 @@ extern "C" void proxy_getSharedData(const char* key_ptr, size_t key_size, const
 extern "C" bool proxy_setSharedData(const char* key_ptr, size_t key_size, const char* value_ptr,
                                     size_t value_size, uint32_t cas);
 
+
+// SharedQueue
+// Note: queue_name must be globally unique.  Registering the same queue_name will overwrite the old registration while
+// preseving any pending data.  Consequently it should typically be followed by a call to proxy_dequeueSharedQueue.
```
PiotrSikora: Why overwrite _and_ preserve old data?

Also, how are extensions expected to guarantee uniqueness when VM can be shared by many of them?

api/wasm/cpp/proxy_wasm_externs.h
PiotrSikora: Also, could we perhaps get `queue_id` upon registration and have `proxy_onSharedQueueReady` use it instead of the string?

api/wasm/cpp/proxy_wasm_externs.h
jplevyak: We need to be able to send messages between VMs so there must be a global namespace, however, we could require that the queue_name be (vm_id, queue_name).  That would prevent collisions and would prevent stealing of someone elses queue.

api/wasm/cpp/proxy_wasm_externs.h
jplevyak: That seems reasonable.  Do you want an "getSharedQueue(vm_id, queue_name)" which would return a queue_id as well for the sender?

PiotrSikora: Would we use that before each call or once during startup? i.e. could the `queue_id` change with new registration?

api/wasm/cpp/proxy_wasm_externs.h
PiotrSikora: Sounds reasonable.

api/wasm/cpp/proxy_wasm_externs.h
jplevyak: Just during setup.

api/wasm/cpp/proxy_wasm_externs.h
jplevyak: We want to preserve the existing data in case there is a reconfigure which causes a VM to restart and we don't want to lose any data.

source/extensions/common/wasm/wasm.cc
```diff
@@ -80,6 +80,47 @@ class SharedData {
...
+    auto it = queues.find(queue_name);
+    if (it == queues.end()) {
+      return false;
+    }
+    it->second.queue.push_back(std::string(value));
+    auto vm_id = it->second.vm_id;
+    auto context_id = it->second.context_id;
+    auto qn = std::string(queue_name);
+    it->second.dispatcher->post([vm_id, context_id, qn] {
+      auto wasm = getThreadLocalWasmOrNull(vm_id);
```
bianpengyuan: This means vm_id cannot be empty when using queue? There will only be one consumer on this queue right? Can't we just keep a pointer to context of that consumer VM?

source/extensions/common/wasm/wasm.cc
```diff
@@ -635,6 +707,25 @@ bool Context::setSharedData(absl::string_view key, absl::string_view value, uint
   return global_shared_data.set(wasm_->id(), key, value, cas);
 }
 
+// Shared Queue
+
+void Context::registerSharedQueue(absl::string_view queue_name) {
+  auto root_context_id = id(); // Assime that Context is a root context.
```
bianpengyuan: s/Assime/Assume

source/extensions/common/wasm/wasm.cc
```diff
@@ -635,6 +707,25 @@ bool Context::setSharedData(absl::string_view key, absl::string_view value, uint
...
+  if (root_context_id_ != 0) { // Root contexts have root_context_id_ == 0.
+    root_context_id = root_context_id_;
```
bianpengyuan: Can we add some comment here such as "convert context of queue registration call from stream context to its root context". It is not obvious on what this two lines do.

jplevyak:
I have updated the interface as discussed.  Both register and resolve now return a queue token.  The queue name is now local to a vm_id, so the names are not "global" anymore.  PTAL

source/extensions/common/wasm/wasm.cc

source/extensions/common/wasm/wasm.cc

source/extensions/common/wasm/wasm.cc
jplevyak: The context can disappear at any time, so we have to look it up.  The empty string is acceptable for the vm_id if you don't want to define one.

test/extensions/filters/http/wasm/wasm_filter_test.cc
```diff
@@ -52,6 +52,18 @@ class TestFilter : public Envoy::Extensions::Common::Wasm::Context {
       : Envoy::Extensions::Common::Wasm::Context(wasm, root_context_id) {}
 
   void scriptLog(spdlog::level::level_enum level, absl::string_view message) override {
+    std::cerr << std::string(message) << "\n";
```
PiotrSikora: Leftover?

test/extensions/filters/http/wasm/wasm_filter_test.cc
```diff
@@ -52,6 +52,18 @@ class TestFilter : public Envoy::Extensions::Common::Wasm::Context {
...
+  }
+  MOCK_METHOD2(scriptLog_, void(spdlog::level::level_enum level, absl::string_view message));
+};
+
+class TestRoot : public Envoy::Extensions::Common::Wasm::Context {
+public:
+  TestRoot() {}
+
+  void scriptLog(spdlog::level::level_enum level, absl::string_view message) override {
+    std::cerr << std::string(message) << "\n";
```
PiotrSikora: Leftover?

test/extensions/filters/http/wasm/wasm_filter_test.cc
```diff
@@ -336,6 +354,40 @@ TEST_F(WasmHttpFilterTest, NullVmPluginRequestHeadersOnly) {
   filter_->onDestroy();
 }
 
+TEST_P(WasmHttpFilterTest, Shared) {
```
PiotrSikora: Nit: `SharedData`.

test/extensions/filters/http/wasm/wasm_filter_test.cc
```diff
@@ -336,6 +354,40 @@ TEST_F(WasmHttpFilterTest, NullVmPluginRequestHeadersOnly) {
...
+  EXPECT_CALL(*filter_,
+              scriptLog_(spdlog::level::warn, Eq(absl::string_view("get 2 shared_data_value2"))));
+
+  Http::TestHeaderMapImpl request_headers{{":path", "/"}};
+  EXPECT_EQ(Http::FilterHeadersStatus::Continue, filter_->decodeHeaders(request_headers, true));
+  StreamInfo::MockStreamInfo log_stream_info;
+  filter_->log(&request_headers, nullptr, nullptr, log_stream_info);
+}
+
+TEST_P(WasmHttpFilterTest, Queue) {
```
PiotrSikora: Nit: `SharedQueue`.

test/extensions/filters/http/wasm/wasm_filter_test.cc
jplevyak: Done

test/extensions/filters/http/wasm/wasm_filter_test.cc
jplevyak: Done

test/extensions/filters/http/wasm/wasm_filter_test.cc
jplevyak: Removed both.

api/wasm/cpp/proxy_wasm_externs.h
```diff
@@ -85,6 +86,17 @@ extern "C" void proxy_getSharedData(const char* key_ptr, size_t key_size, const
 extern "C" bool proxy_setSharedData(const char* key_ptr, size_t key_size, const char* value_ptr,
                                     size_t value_size, uint32_t cas);
 
+
```
PiotrSikora: Nit: 2 empty lines.

source/extensions/common/wasm/wasm.cc
```diff
@@ -94,11 +152,39 @@ class SharedData {
   }
 
 private:
+  uint32_t nextQueueToken() {
+    while (true) {
+      uint32_t token = next_queue_token++;
+      if (!token) {
```
PiotrSikora: Nit: since this is a number, `token == 0`.

source/extensions/common/wasm/wasm.cc
```diff
@@ -640,6 +762,30 @@ bool Context::setSharedData(absl::string_view key, absl::string_view value, uint
   return global_shared_data.set(wasm_->id(), key, value, cas);
 }
 
+// Shared Queue
+
+uint32_t Context::registerSharedQueue(absl::string_view queue_name) {
+  auto root_context_id = id(); // Assume that Context is a root context.
+  if (root_context_id_ != 0) { // Root contexts have root_context_id_ == 0.
```
PiotrSikora:
```
if (root_context_id_ == 0) {
    // Context is a root context.
    root_context_id = id();
} else {
    // Convert context from stream context to root context.
    root_context_id = root_context_id_;
}
```
would avoid extra assignment, and is IMHO slightly more readable.

jplevyak: The compiler will eliminate the extra assignment.  What about ? :

api/wasm/cpp/proxy_wasm_externs.h
jplevyak: Done.

source/extensions/common/wasm/wasm.cc

source/extensions/common/wasm/wasm.cc

jplevyak:
Fixed the nit.

PR #51, opened 2019-06-12
build: update dependencies in build containers.

PiotrSikora:
While there, revert local changes to reduce the diff with upstream.

### DIFF
--- a/ci/build_container/build_container_ubuntu.sh
+++ b/ci/build_container/build_container_ubuntu.sh
@@ -5,18 +5,11 @@ set -e
 # Setup basic requirements and install them.
 apt-get update
 export DEBIAN_FRONTEND=noninteractive
-apt-get install -y curl wget software-properties-common make cmake git python python-pip python3 python3-pip \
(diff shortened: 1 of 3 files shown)


ci/build_container/build_container_ubuntu.sh
```diff
@@ -31,7 +24,7 @@ update-alternatives --config gcc
 update-alternatives --config g++
 update-alternatives --config gcov
 # Bazel and related dependencies.
-apt-get install -y openjdk-8-jdk
+apt-get install -y openjdk-8-jdk curl
```
jplevyak: Why are these not included in the first list above?  I noticed that you shifted to wget for the llvm key because curl is no longer available here.

PiotrSikora: This is part of "revert local changes to reduce diff with upstream".  I originally changed it when adding CMake in 62d70b228c4f4969bd868ced1b34b9d48873d82d, but it was unnecessary "cleanup" that I never upstreamed.

The only change in this file (compared to upstream) is addition of `libglib2.0-dev`.

Sorry for making diffs hard...

PR #53, opened 2019-06-14
 ci: configure CircleCI for Envoy-WASM.

PiotrSikora:


PiotrSikora:
Includes https://github.com/envoyproxy/envoy/pull/7269.

PiotrSikora:
@mattklein123 this should be good to go, if you can enable CircleCI.

mattklein123:
@PiotrSikora should be enabled. If you kick this I think it should build.

PiotrSikora:
It looks that WAVM is crashing on CircleCI in dev build (but it passes in my local environment).

@jplevyak any ideas from the stacktrace?

PiotrSikora:
Sigh, I think we're synced at a bad commit from upstream, since `gcovr` pulls bunch of interfaces from and generated protobuf files for hot restart, even for non-WASM tests, which isn't the case in upstream's master, which triggers a bug in `gcovr` when it cannot find a common patch.

This should be fixed now.

lizan:
@PiotrSikora
That should be fixed in coverage script, genfiles are now merged with bin directory. https://github.com/envoyproxy/envoy/pull/6866/commits/94667d870159ecd64388686d389d1576a88cf51f

PiotrSikora:
@lizan thanks!

PR #54, opened 2019-06-14
Fix inittialization order issue.

jplevyak:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


PiotrSikora:
Why are you removing examples? If it's because of `check_spelling`, then you can exclude `examples/wasm/envoy_filter_http_wasm_example.wat` by adding it to `tools/spelling_skip_files.txt`.

jplevyak:
re-added those files and added them to tools/spelling_skip_files.txt

tools/spelling_skip_files.txt
```diff
@@ -1,2 +1,4 @@
 OWNERS.md
 test/extensions/filters/http/wasm/test_data/metadata_cpp.wat
+examples/wasm/envoy_filter_http_wasm_example.wat
+examples/wasm/envoy_filter_http_wasm_example.wasm
```
PiotrSikora: Nit: `check_spelling` complains only about the `.wat` file.

Side-note: we could probably just use a `*.wat` regex here to future-proof it.

jplevyak: Done.

PR #55, opened 2019-06-15
ci: enable WAVM.

PiotrSikora:
Signed-off-by: Piotr Sikora <piotrsikora@google.com>

PR #57, opened 2019-06-15
ci: run tests on macOS.

PiotrSikora:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,13 +77,13 @@ jobs:
        - checkout
        - restore_cache:
            keys:
-             - macos-{{ checksum "bazel/repository_locations.bzl" }}
+             - macos-cache-{{ checksum "bazel/repository_locations.bzl" }}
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -65,6 +65,26 @@ jobs:
        # TODO(PiotrSikora): re-enable once fixed.
        # - run: ci/do_circle_ci.sh check_spelling_pedantic
 
+   macos:
+     macos:
+       xcode: "10.2.1"
+     environment:
+       BAZEL_BUILD_EXTRA_OPTIONS: "--define wasm=v8" # v8 only, WAVM segfaults in tests.
```
jplevyak: In release and debug ?

PiotrSikora: Nn `fastbuild`, which is what `ci/mac_ci_steps.sh` runs.

I have separate PR for investigating WAVM builds (#55), so let's not block this one on it.

PR #72, opened 2019-07-16
2 initial stress tests for the http wasm filters

duderino:
@jplevyak this should get you started and I'll keep adding additional test cases.  

Test it like so:

```
bazel test --define wasm=enabled  //test/extensions/filters/http/wasm:wasm_filter_stress_test 
```

there's a basic grpc callout test that mostly works... except the wasm filter thinks the grpc response is bad.  It logs this for every grpc response it receives:

```
[2019-07-16 04:36:18.547][55][info][wasm] [source/extensions/common/wasm/wasm.cc:1272] wasm log: failure 2
```

The http callout test throws an exception inside the wasm bundle.  that one is currently skipped.


api/wasm/cpp/proxy_wasm_impl.h
```diff
@@ -288,7 +288,7 @@ class ContextBase {
 class RootContext : public ContextBase {
 public:
   RootContext(uint32_t id, StringView root_id) : ContextBase(id), root_id_(root_id) {}
-  ~RootContext() {}
+  virtual ~RootContext() {}
```
duderino: ContextBase's dtor is virtual so I thought I'd throw this in there

source/extensions/filters/http/wasm/config.cc
```diff
@@ -28,9 +28,7 @@ Http::FilterFactoryCb WasmFilterConfig::createFilterFactoryFromProtoTyped(
 /**
  * Static registration for the Wasm filter. @see RegisterFactory.
  */
-static Registry::RegisterFactory<WasmFilterConfig,
-                                 Server::Configuration::NamedHttpFilterConfigFactory>
-    register_;
+REGISTER_FACTORY(WasmFilterConfig, Server::Configuration::NamedHttpFilterConfigFactory);
```
duderino: And all the other filters used this REGISTER_FACTORY macro so added that too

test/extensions/filters/http/wasm/test_data/grpc_callout_cpp.cc
```diff
@@ -0,0 +1,51 @@
...
+    logInfo(response.string_value());
+    continueRequest();
+  }
+  void onFailure(GrpcStatus status,
+                 std::unique_ptr<WasmData> error_message) override {
+    logInfo(std::string("failure ") + std::to_string(static_cast<int>(status)) +
+            std::string(error_message->view()));
+
+    // TODO wasm engine must support fail closed: expose abortRequest() or similar
+    continueRequest();
```
duderino: This test case passes because of this continueRequest().  There should be some mechanism to abort the downstream request onFailure.

test/extensions/filters/http/wasm/test_data/http_callout_cpp.cc
```diff
@@ -0,0 +1,50 @@
...
+    WasmDataPtr data = getRequestHeader("x-callout-url");
+
+    if (!data) {
+      logWarn("Missing x-callout-url header, cannot forward");
+      return FilterHeadersStatus::Continue;
+    }
+
+    StringView callout_url{data->toString()};
+
+    logInfo("Forwarding to: " + std::string(callout_url));
```
duderino: @jplevyak take note.  After 20 or so iterations the callout_url will log something that looks corrupted

test/extensions/filters/http/wasm/wasm_filter_stress_test.cc
```diff
@@ -0,0 +1,234 @@
...
+#elif defined(ENVOY_WASM_V8)
+                                              "envoy.wasm.vm.v8"
+#elif defined(ENVOY_WASM_WAVM)
+                                              "envoy.wasm.vm.wavm"
+#endif
+                                              ),
+                                          testing::Values("cpp")));
+
+TEST_P(WasmStressTest, GrpcCalloutFailOpenHappyPath) {
+  Envoy::Logger::Registry::setLogLevel(spdlog::level::info);
```
duderino: Does anyone know how to set this on the bazel commandline?

test/extensions/filters/http/wasm/wasm_filter_stress_test.cc
```diff
@@ -0,0 +1,234 @@
...
+  EXPECT_EQ(client->class4xxResponses(), 0);
+  EXPECT_EQ(client->class5xxResponses(), 0);
+  EXPECT_EQ(client->responseTimeouts(), 0);
+  // No client sockets are rudely closed by server / no client sockets are
+  // reset.
+  EXPECT_EQ(client->remoteCloses(), 0);
+
+  // assert that the callout server and origin server see every request
+  EXPECT_EQ(clusterHelper(StressTest::ORIGIN_CLUSTER_NAME).requestsReceived(), requests_to_send);
+  EXPECT_EQ(clusterHelper(callout_cluster_name).requestsReceived(), requests_to_send);
```
duderino: Assertions here can be built out once the gRPC callout response is handled correctly.  Counters can also be dumped and asserted on (see StressTest superclass)

test/extensions/filters/http/wasm/wasm_filter_stress_test.cc
```diff
@@ -0,0 +1,234 @@
...
+}
+
+TEST_P(WasmStressTest, HttpCalloutFailOpenHappyPath) {
+  // TODO fix test. Currently fails with a:
+  // terminate called after throwing an instance of 'Envoy::Extensions::Common::Wasm::WasmException'
+  //  what():  emscripten cxa_allocate_exception
+  GTEST_SKIP();
```
duderino: Skipped because of the unhandled exception inside the http callout bundle

test/integration/integration.cc
```diff
@@ -294,10 +294,9 @@ void BaseIntegrationTest::createUpstreams() {
 }
 
 void BaseIntegrationTest::createEnvoy() {
-  std::vector<uint32_t> ports;
   for (auto& upstream : fake_upstreams_) {
     if (upstream->localAddress()->ip()) {
-      ports.push_back(upstream->localAddress()->ip()->port());
+      ports_.push_back(upstream->localAddress()->ip()->port());
```
duderino: the StressTest subclass needs to set these ports without creating FakeUpstreams

test/integration/integration.h
```diff
@@ -135,7 +135,7 @@ struct ApiFilesystemConfig {
 /**
  * Test fixture for all integration tests.
  */
-class BaseIntegrationTest : Logger::Loggable<Logger::Id::testing> {
+class BaseIntegrationTest : protected Logger::Loggable<Logger::Id::testing> {
```
duderino: lets the StressTest subclass use `ENVOY_LOG`

test/extensions/filters/http/wasm/wasm_filter_stress_test.cc
PiotrSikora: Try `bazel test --define wasm=enabled //test/... -- -l info`

duderino: Thanks @PiotrSikora

api/wasm/cpp/proxy_wasm_impl.h
jplevyak: Sure for documentation.  It is virtual anyway as per the spec.

test/extensions/filters/http/wasm/test_data/grpc_callout_cpp.cc
jplevyak: Do we have an issue for this?

test/extensions/filters/http/wasm/test_data/grpc_callout_cpp.cc
```diff
@@ -0,0 +1,51 @@
...
+};
+
+static RegisterContextFactory
+    register_ExampleContext(CONTEXT_FACTORY(ExampleContext));
+
+class MyGrpcCallHandler : public GrpcCallHandler<google::protobuf::Value> {
+ public:
+  MyGrpcCallHandler(ExampleContext *context)
+      : GrpcCallHandler<google::protobuf::Value>(context), context_(context) {}
+  void onCreateInitialMetadata() override {}
```
jplevyak: Should we convert these from pure virtuals to having a default implementation of {} so that we can avoid requiring handlers which are used?

test/stress/stress_test.cc
```diff
@@ -0,0 +1,58 @@
+#include "stress_test.h"
+
+namespace Envoy {
+namespace Stress {
+
+const std::string StressTest::ORIGIN_CLUSTER_NAME{"origin_cluster"};
+
+// TODO dump shows codec type is still HTTP1
```
jplevyak: So the clients are generating HTTP1?

test/extensions/filters/http/wasm/test_data/grpc_callout_cpp.cc
duderino: Sure, would you like me to do that in this PR?

test/extensions/filters/http/wasm/test_data/grpc_callout_cpp.cc
duderino: https://github.com/envoyproxy/envoy-wasm/issues/74

test/stress/stress_test.cc
duderino: Not sure.  Proposal:  next I'll add additional parameters to the test case so we can cross with http vs h2 and IPv4 vs. IPv6.   I'll make sure the right thing is being sent on the wire.

When I'm done I'll update this PR if it's still open, else I'll open a new one

jplevyak:
It seems that exception support in emscripten is... poor.  First, in order to get exception support we need to add "-s DISABLE_EXCEPTION_CATCHING=0" to the Makefile.

After that, we would need  a great deal of support, implementing a raft of invoke_XXX and dynCall_XXX functions in addition to ___cxa_ calls.

All this is likely to change as well when we move to the new upstream (LLVM) based backend for emscripten.

The upshot is that I think we should drop exception support within the VM until this gets sorted out.

Thoughts?

duderino:
> It seems that exception support in emscripten is... poor. First, in order to get exception support we need to add "-s DISABLE_EXCEPTION_CATCHING=0" to the Makefile.
> 
> After that, we would need a great deal of support, implementing a raft of invoke_XXX and dynCall_XXX functions in addition to ___cxa_ calls.
> 
> All this is likely to change as well when we move to the new upstream (LLVM) based backend for emscripten.
> 
> The upshot is that I think we should drop exception support within the VM until this gets sorted out.
> 
> Thoughts?

@jplevyak if we drop exception support, what happens if third party code, etc in a wasm bundle throws an exception?

test/stress/stress_test.cc
duderino: I turned this into a parameterized test that uses HTTP1 vs. HTTP2 X IPv4 vs. IPv6.    In the process I addressed this TODO

test/extensions/filters/http/wasm/wasm_filter_stress_test.cc
```diff
@@ -0,0 +1,282 @@
...
+                                          testing::Values("IPv4", "IPv6")));
+
+TEST_P(GrpcWasmStressTest, CalloutHappyPath) {
+  constexpr uint32_t connections_to_initiate = 30;
+  constexpr uint32_t requests_to_send = 30 * connections_to_initiate;
+  const std::string wasm_file = absl::StrCat(
+      TestEnvironment::runfilesDirectory(),
+      "/test/extensions/filters/http/wasm/test_data/grpc_callout_", wasmLang(), ".wasm");
+  // Must match cluster name in the wasm bundle:
+  const std::string callout_cluster_name{"callout_cluster"};
```
duderino: @mandarjog Here is an example wasm test case.  The link between this test case and the wasm bundle is this 'callout_cluster' name.  

This test case creates two clusters:  a 'callout_cluster' to simulate some random upstream that the wasm bundle can callout to and a 'origin_cluster' which the data plane request will be forwarded to.    The grpc_callout_cpp.cc file passes the 'callout_cluster' name to the grpc library and Envoy does the rest

test/extensions/filters/http/wasm/test_data/http_callout_cpp.cc
jplevyak: It looks like this is disable for the exception issue.

test/extensions/filters/http/wasm/test_data/grpc_callout_cpp.cc
jplevyak: Nothing, just wondering if you thought that was a good idea.

duderino: Oh, I'm pretty neutral on that.  Most people will copy the boilerplate anyways and then hack it up.

silentdai:
@duderino Could you split the PR? Even we exclude the generated code this PR is no doubt a huge one. Very hard to review

duderino:
> @duderino Could you split the PR? Even we exclude the generated code this PR is no doubt a huge one. Very hard to review

The test framework was already reviewed by @jplevyak back in the istio/proxy repo, so I though we could just not review the framework part (everything under stress/)?

Once this is merged and has a few additional test cases I'm planning on sending the framework to envoyproxy/envoy where it'll experience another batch of reviews

duderino:
Closing in favor of the split in #89 and #90

PR #73, opened 2019-07-16
Add several data accessors into wasm API

bianpengyuan:
Add `getHeaderMapSize`, `getDestinationPort` and `getResponseCode` into WASM API.

cc @jplevyak @PiotrSikora


bianpengyuan:
It is for the request size metrics that we collect: https://cloud.google.com/monitoring/api/metrics_anthos#anthos-istio.io

PiotrSikora:
Erm, `getHeaderMapSize()` has nothing to do with sent / received bytes as needed in metrics.

bianpengyuan:
@PiotrSikora Hmm, but there is no good way to get the bytes transferred on wire easily in filter right? HeaderMapSize() + body_buffer_size could give us a rough idea on the bytes transferred.

PiotrSikora:
@bianpengyuan it looks that `StreamInfo` has `bytesReceived()`... but it returns only body bytes received, without headers... so indeed, there is no easy way to do that right now.

But exporting `HeaderMapSize() + body_buffer_size` as `received_bytes_count ` is wrong (fine for the demo, though), especially with HTTP/2, because of HTTP/2 framing overhead and HPACK, where the number of bytes calculated this way can be an order of magnitude different from what's on the wire. This also doesn't include the TLS/SSL overhead.

bianpengyuan:
@PiotrSikora Ok, looks like the change needs to be made in upstream.

Issue #74, opened 2019-07-16
WASM http filters should be able to send an error to the downstream client after return from a callout


duderino:
This would be necessary to re-implement the istio/proxy mixer filter as a wasm bundle.  AFAIK, it's possible to suspend a client request during the callout and resume it after the callout finishes, but I don't know if it's possible to reject the client request if the callout finishes with an error.

example:

```
Skip to content
 
Search or jump to…

Pull requests
Issues
Marketplace
Explore
 
@duderino 
19
34 9 envoyproxy/envoy-wasm 
 Code  Issues 19  Pull requests 6
 ZenHub  Projects 0  Wiki  Security  Insights
envoy-wasm/test/extensions/filters/http/wasm/test_data/grpc_callout_cpp.cc
@duderino duderino Move wasm filter stress tests to test/extensions/filters/http/wasm.
753eb0e 12 hours ago
52 lines (42 sloc)  1.69 KB
    
// NOLINT(namespace-envoy)
#include <string>
#include <unordered_map>
#include "proxy_wasm_intrinsics.h"
#include "proxy_wasm_intrinsics_lite.pb.h"

class ExampleContext : public Context {
 public:
  explicit ExampleContext(uint32_t id, RootContext *root) : Context(id, root) {}

  FilterHeadersStatus onRequestHeaders() override;
};

static RegisterContextFactory
    register_ExampleContext(CONTEXT_FACTORY(ExampleContext));

class MyGrpcCallHandler : public GrpcCallHandler<google::protobuf::Value> {
 public:
  MyGrpcCallHandler(ExampleContext *context)
      : GrpcCallHandler<google::protobuf::Value>(context), context_(context) {}
  void onCreateInitialMetadata() override {}
  void onSuccess(google::protobuf::Value &&response) override {
    logInfo(response.string_value());
    continueRequest();
  }
  void onFailure(GrpcStatus status,
                 std::unique_ptr<WasmData> error_message) override {
    logInfo(std::string("failure ") + std::to_string(static_cast<int>(status)) +
            std::string(error_message->view()));

   HERE IS WHERE ONE WOULD WANT TO SEND AN ERROR RESPONSE TO THE DOWNSTREAM CLIENT AND ABORT THE TRANSACTION
  }

 private:
  ExampleContext *context_;
};

FilterHeadersStatus ExampleContext::onRequestHeaders() {
  GrpcService grpc_service;
[...]

PR #79, opened 2019-07-18
Add support for responding locally (e.g with an error).

jplevyak:
Added support for clearing the route cache.
Converted GrpcStatus to use the envoy names.

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


jplevyak:
https://github.com/envoyproxy/envoy-wasm/issues/74

jplevyak:
I rebuilt the .wasm files with the latest emscripten, and the tests worked for me locally.  It seems like there is some issue with the magic numbers in the module.

source/extensions/common/wasm/null/wasm_api_impl.h
```diff
@@ -73,6 +73,17 @@ inline void proxy_setMetadataStruct(MetadataType type, const char* name_ptr, siz
 // Continue
 inline void proxy_continueRequest() { continueRequestHandler(current_context_); }
 inline void proxy_continueResponse() { continueResponseHandler(current_context_); }
+inline void proxy_sendLocalReply(uint32_t response_code, const char* response_code_details_ptr,
```
PiotrSikora: Nit: I'd prefer if this was called `sendLocalResponse`.

test/extensions/wasm/wasm_test.cc
```diff
@@ -175,9 +175,9 @@ TEST_P(WasmTest, EmscriptenVersion) {
   uint32_t major = 9, minor = 9, abi_major = 9, abi_minor = 9;
   EXPECT_TRUE(wasm->getEmscriptenVersion(&major, &minor, &abi_major, &abi_minor));
   EXPECT_EQ(major, 0);
-  EXPECT_EQ(minor, 0);
+  EXPECT_TRUE(minor >= 0 && minor <= 1);
   EXPECT_EQ(abi_major, 0);
-  EXPECT_EQ(abi_minor, 1);
+  EXPECT_TRUE(abi_minor >= 0 && abi_minor <= 3);
```
PiotrSikora: Unrelated changes?

source/extensions/common/wasm/null/wasm_api_impl.h
jplevyak: I did this to match the envoy name, but I agree, that is a better name.

test/extensions/wasm/wasm_test.cc
jplevyak: The provides support for up to 1.38.39 of emscripten.  Yes, it is unrelated.  I can pull it out if you want.

jplevyak:
The asan test fails with a timeout and

tcpdump: No such file or directory

source/extensions/common/wasm/null/wasm_api_impl.h
PiotrSikora: We don't match Envoy names for most part (e.g. request/response vs encode/decode), so I don't see a strong need to do it here either.

test/extensions/wasm/wasm_test.cc
PiotrSikora: I'd prefer to have separate PRs for working features that clearly describe intention behind the change, than sneaked in changes and/or code dumps.

PiotrSikora:
Yeah, asan has been timing out from time to time lately.

PiotrSikora:
Can you merge master? Thanks!

source/extensions/common/wasm/null/wasm_api_impl.h
jplevyak: Changed.  I almost just called it "respond" but that is a bit terse.

test/extensions/wasm/wasm_test.cc
jplevyak: OK, I'll pull that out.

jplevyak: https://github.com/envoyproxy/envoy-wasm/pull/83

jplevyak:
updated to master

api/wasm/cpp/proxy_wasm_intrinsics.h
```diff
@@ -10,7 +10,7 @@
 #endif
 
 #include <cstdint>
-using EnumType = uint32_t;
+using EnumType = int32_t;
```
silentdai: Why are you making this change? For the Invalid as "-1"? You could have replace it by "std::numeric_limits<EnumType>::max()"

api/wasm/cpp/proxy_wasm_impl.h
```diff
@@ -631,9 +631,19 @@ inline google::protobuf::Struct Context::getResponseMetadataStruct(StringView na
   return getMetadataStruct(MetadataType::Response, name);
 }
 
-// Continue
+// Continue/Respond/Route
 inline void continueRequest() { proxy_continueRequest(); }
 inline void continueResponse() { proxy_continueResponse(); }
+inline void sendLocalResponse(uint32_t response_code, StringView response_code_details, StringView body,
+    const HeaderStringPairs& additional_response_headers, uint32_t grpc_status = GrpcStatus::InvalidCode) {
```
silentdai: there is an implicit int32_t to uint32_t. See my suggestion in EnumType

source/extensions/common/wasm/wasm.cc
```diff
@@ -431,6 +431,36 @@ void continueResponseHandler(void* raw_context) {
...
+                              Word response_code_details_size, Word body_ptr, Word body_size,
+                              Word additional_response_header_pairs_ptr,
+                              Word additional_response_header_pairs_size, Word grpc_code) {
+  auto context = WASM_CONTEXT(raw_context);
+  auto details =
+      context->wasmVm()->getMemory(response_code_details_ptr, response_code_details_size);
+  auto body = context->wasmVm()->getMemory(body_ptr, body_size);
+  auto additional_headers = toPairs(context->wasmVm()->getMemory(
+      additional_response_header_pairs_ptr, additional_response_header_pairs_size));
+  std::function<void(Http::HeaderMap & headers)> modify_headers =
```
silentdai: is this formatted?

silentdai: use auto to obtain lambda type. lambda will convert to function when it is needed

source/extensions/common/wasm/wasm.cc
```diff
@@ -1650,6 +1684,7 @@ void Wasm::registerCallbacks() {
     _REGISTER(___cxa_throw);
     _REGISTER(___cxa_pure_virtual);
     _REGISTER(___cxa_allocate_exception);
+    _REGISTER(___cxa_uncaught_exception);
```
silentdai: For my curiosity: could you explain why we need it here? Link is also good. Thanks!

source/extensions/common/wasm/null/null.h
```diff
@@ -15,7 +15,7 @@ using FilterHeadersStatus = Http::FilterHeadersStatus;
 using FilterMetadataStatus = Http::FilterMetadataStatus;
 using FilterTrailersStatus = Http::FilterTrailersStatus;
 using FilterDataStatus = Http::FilterDataStatus;
-using GrpcStatus = ProtobufUtil::error::Code;
+using GrpcStatus = Envoy::Grpc::Status::GrpcStatus;
```
silentdai: Naive question: why?

api/wasm/cpp/proxy_wasm_intrinsics.h
jplevyak: I am matching the envoy GrpcStatus class so we can use that one when using the null VM.

source/extensions/common/wasm/null/null.h
jplevyak: Because the latter defines an invalid status which I can use to indicate that the gRPC status is not provided because the envoy function takes std::optional<GrpcStatus> but the ABI is based on raw words, so std::optional is not available and I need a sentinel value.

source/extensions/common/wasm/wasm.cc
jplevyak: This is to support emscripten 1.38.39.  This will disappear from this PR when #83 is merged.

jplevyak:
I am going to hold off merging till #83 is merged.

api/wasm/cpp/proxy_wasm_enums.h
```diff
@@ -34,24 +34,25 @@ enum class HeaderMapType : EnumType {
   GrpcReceiveTrailingMetadata = 6,  // Immutable
 };
 enum GrpcStatus : EnumType {
```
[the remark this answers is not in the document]
jplevyak: done

api/wasm/cpp/proxy_wasm_impl.h
jplevyak: Made the conversion explicit.

source/extensions/common/wasm/wasm.cc
jplevyak: yes, but now it is changed as per below.

jplevyak: done

silentdai:
/lgtm

Thanks!

PR #99, opened 2019-07-26
Return error codes for WebAssembly metadata operations.

jplevyak:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


jplevyak:
@mandarjog FYI

source/extensions/common/wasm/null/null.h
```diff
@@ -18,7 +18,7 @@ using FilterDataStatus = Http::FilterDataStatus;
 using GrpcStatus = Envoy::Grpc::Status::GrpcStatus;
 using MetricType = Envoy::Extensions::Common::Wasm::Context::MetricType;
 using StringView = absl::string_view;
-using EnumType = uint32_t;
```
jplevyak: FYI: this is to be compatible with envoy.

api/wasm/cpp/proxy_wasm_enums.h
```diff
@@ -59,3 +59,11 @@ enum class MetricType : EnumType {
   Gauge = 1,
   Histogram = 2,
 };
+enum class MetadataResult : EnumType {
```
PiotrSikora: This is fine, but ultimately I think that we need to have `WasmResult` (or something like that) that we can use across the whole ABI surface, instead of having `MetadataResult`, `HttpResult`, etc. Similar to how syscalls share the error codes.

What do you think?

jplevyak: Yes, that seems reasonable.  Let's create an issue and do that in another PR.  There are other cases which we can roll into that PR.

jplevyak: https://github.com/envoyproxy/envoy-wasm/issues/104

PR #105, opened 2019-07-30
gRPC and HTTP calls return on RootContext.  Add setEffectiveContext()…

jplevyak:
… call.

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


jplevyak:
@mandarjog @bianpengyuan 

We want gRPC and HTTP calls to come back on the RootContext in case the stream disappears. 
 In order to support "continue", use setEffectiveContext().

This also enables de-duping of calls to the control plane by using a single outstanding request for multiple waiting streams to prevent head-of-line (stampede) issues.

api/wasm/cpp/proxy_wasm_impl.h
```diff
@@ -147,26 +147,28 @@ using HeaderStringPairs = std::vector<std::pair<std::string, std::string>>;
 
 class GrpcCallHandlerBase {
 public:
-  GrpcCallHandlerBase(Context* context) : context_(context) {}
+  GrpcCallHandlerBase() {}
   virtual ~GrpcCallHandlerBase() {}
 
+  RootContext* context() { return context_; }
```
bianpengyuan: Why expose this method?

source/extensions/common/wasm/wasm.cc
```diff
@@ -328,6 +332,18 @@ const uint8_t* decodeVarint(const uint8_t* pos, const uint8_t* end, uint32_t* ou
...
 
+Context* ContextOrEffectiveContext(Context* context) {
+  if (effective_context_id_ == 0) {
+    return context;
+  }
+  auto effective_context = context->wasm()->getContext(effective_context_id_);
+  if (effective_context) {
+    return effective_context;
+  }
+  // The effective_ccontext_id_ no longer exists, revert to the true context.
```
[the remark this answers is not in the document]
jplevyak: done.

api/wasm/cpp/proxy_wasm_impl.h
jplevyak: If we go with your suggestion of automatically setting the effective context, then both the root context and the stream context will be available from the HandlerBase.   Why not make those available to the user so that they don't have to store extra parameters? Let me make the changes as per your suggestion and see what it looks like.

jplevyak: What about the case of httpCall and grpcSimpleCall.  These take simple functions and so would have to store the context_id_ and use getContext(context_id_) to probe if the stream context still exists to determine which effective context they have.  Wouldn't it be easier to always know that the effective context is the RootContext ?  For the grpcCallHandler and grpcStreamHandler situations we have an object we can hang a more complex API over.  We could do as you suggest in that case and provide two functions: root_context() and context() where the latter would return nullptr if the context had been deleted.  We could set the effective context to be the stream context if it exists and it would make more sense because we are providing an API which returns the context (or null).   But then it is inconsistent with the pure std::function API.    Thoughts?

bianpengyuan: Yeah this is a good point. I don't like inconsistency either. I am thinking of some deeper change, but I don't think it is worth to do now. I am thinking should we consolidate all call out functions to use handler? Like for simple call functions, a handler is created inside the call, which means `grpc_simple_calls` map could be removed and http call could be handled by a handler object instead of just a callback function. Does this make sense?

This is a rather big refactor considering the purpose of this PR and should be something to consider when upstream if change makes sense. So I am OK with the current implementation.

bianpengyuan: Yeah this is a good point. I don't like inconsistency either. I am thinking of some deeper change, but I don't think it is worth to do now. I am thinking should we consolidate all call out functions to use handler? Like for simple call functions, a default simple handler is created inside the call, which means grpc_simple_calls map could be removed and a http call handler class needs to be added. Does this make sense?

This is a rather big refactor considering the purpose of this PR and should be something to consider when upstream if change makes sense. So I am OK with the current implementation.

jplevyak: https://github.com/envoyproxy/envoy-wasm/issues/111

Created.   Let's move the re-design discussion there.

PR #112, opened 2019-08-02
Make field separator and value separator configurable.

mandarjog:
The metrics subsystem flattens metric dimensions using `.`
However some values like destination service has `.` in them.
This causes the tag extractor to break.

We need a way to configure these 2 fields using the Metric object to make tag extraction correct.

### DIFF
--- a/api/wasm/cpp/proxy_wasm_impl.h
+++ b/api/wasm/cpp/proxy_wasm_impl.h
@@ -997,7 +997,9 @@ struct MetricTag {
 
 struct MetricBase {
   MetricBase(MetricType t, const std::string& n) : type(t), name(n) {}
-  MetricBase(MetricType t, const std::string& n, std::vector<MetricTag> ts, std::string  fs=".", std::string vs=".")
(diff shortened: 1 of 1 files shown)


api/wasm/cpp/proxy_wasm_impl.h
```diff
@@ -997,15 +997,18 @@ struct MetricTag {
 
 struct MetricBase {
   MetricBase(MetricType t, const std::string& n) : type(t), name(n) {}
-  MetricBase(MetricType t, const std::string& n, std::vector<MetricTag> ts)
-      : type(t), name(n), tags(ts.begin(), ts.end()) {}
+  MetricBase(MetricType t, const std::string& n, std::vector<MetricTag> ts, std::string  fs=".", std::string vs=".")
```
jplevyak: Could you remove the default arguments here and add back the version without the additional arguments (i.e. have 2 constructors one without fs and vs and one with).

That way the "." and "," constants appear only in one place (as the default initializers next to the definition of the instance variables).

PR #113, opened 2019-08-02
Add support for Metadata Expressions including. tlsVersion.

jplevyak:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


jplevyak:
@mandarjog

api/wasm/cpp/proxy_wasm_enums.h
```diff
@@ -15,15 +15,29 @@ enum class FilterDataStatus : EnumType {
...
+  Listener = 6,       // Immutable
+  Cluster = 7,        // Immutable
+  Expression = 8,     // The key is a string expression. Only proxy_getMetadata().
 };
+/*
+  Expression and their types:
+
+  request.protocol : string
+  response.protocol : string
+  request.destination_port : int32
```
kyessenov: We should be careful not to define new names for things that are established in https://istio.io/docs/reference/config/policy-and-telemetry/attribute-vocabulary/ or https://cloud.google.com/iam/docs/conditions-overview. I'm happy to make another pass to fix the names in a follow-up

kyessenov:
Just curious: why do we commit the large binary files? is there a CI check validating them?

api/wasm/cpp/proxy_wasm_enums.h
mandarjog: We should just use consistent names now, instead of another PR.

jplevyak: I agree that we should have consistent names when we upstream, but this is an experimental repository and Kuat is going to replace all this code very soon.  This is just to get Mandar the last "must have" interface.  That said, if the correct names are obvious I'd like to make the change in this PR.

jplevyak:
RE: the large binaries.  We have an issue to get emscripten working with bazel and then we can generate these files on the fly.  I agree that it is really annoying.  As soon as we get some cycles I am going to work on it.

api/wasm/cpp/proxy_wasm_enums.h
jplevyak: The current names should not be consisdered part of the external interface.  We have the accessors for that, e.g. getRequestDestinationPort().  I can also prefix the current names with "_experimental_" or "_internal_" which will give a clear indication these are not part of a stable exernal interface.

jplevyak:
Ping.

PR #122, opened 2019-08-09
Add trafic direction API.

jplevyak:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


jplevyak:
@mandarjog

source/extensions/common/wasm/wasm.h
```diff
@@ -1072,6 +1078,12 @@ struct SaveRestoreContext {
   uint32_t saved_effective_context_id_;
 };
 
+inline PluginDirection
+pluginDirectionFromTrafficDirection(envoy::api::v2::core::TrafficDirection direction) {
+  return direction == envoy::api::v2::core::TrafficDirection::INBOUND ? PluginDirection::Inbound
```
kyessenov: We should default unspecified as inbound in istio context (since older control planes don't set the field).

jplevyak: That would mean changing the value from UNSPECIFIED to INBOUND.  Rather I think we should just document in the istio plugin docs that UNSPECIFIED should be interpreted as INBOUND where applicable.

source/extensions/common/wasm/wasm.h
```diff
@@ -844,6 +850,20 @@ inline bool Wasm::copyToPointerSize(const Buffer::Instance& buffer, uint64_t sta
   return true;
 }
 
+inline PluginDirection
+pluginDirectionFromTrafficDirection(envoy::api::v2::core::TrafficDirection direction) {
+  switch (direction) {
+  default:
```
PiotrSikora: Nit: move `default` case to the end.

api/wasm/cpp/proxy_wasm_enums.h
```diff
@@ -37,6 +37,7 @@ enum class MetadataType : EnumType {
   response.response_code : int32
   upstream.tls_version : string
   downstream.tsl_version : string
+  plugin.direction : int32 (0 = INBOUND, 1 = OUTBOUD)
```
PiotrSikora: Is this comment still accurate?

source/extensions/common/wasm/wasm.h

api/wasm/cpp/proxy_wasm_enums.h
jplevyak: Fixed (now refers to the enum below).

source/extensions/common/wasm/wasm.h
```diff
@@ -844,6 +850,20 @@ inline bool Wasm::copyToPointerSize(const Buffer::Instance& buffer, uint64_t sta
...
+pluginDirectionFromTrafficDirection(envoy::api::v2::core::TrafficDirection direction) {
+  switch (direction) {
+  case envoy::api::v2::core::TrafficDirection::UNSPECIFIED:
+    return PluginDirection::Unspecified;
+  case envoy::api::v2::core::TrafficDirection::INBOUND:
+    return PluginDirection::Inbound;
+  case envoy::api::v2::core::TrafficDirection::OUTBOUND:
+    return PluginDirection::Outbound;
+  default:
+    ASSERT(!"Bad envoy::api::v2::core::TrafficDirection");
```
PiotrSikora: Bah, this won't work in release builds:
```
In file included from bazel-out/k8-opt/bin/source/extensions/access_loggers/wasm/_virtual_includes/wasm_access_log_lib/extensions/access_loggers/wasm/wasm_access_log_impl.h:8:
bazel-out/k8-opt/bin/source/extensions/common/wasm/_virtual_includes/wasm_hdr/extensions/common/wasm/wasm.h:865:1: error: control may reach end of non-void function [-Werror,-Wreturn-type]
}
```

source/extensions/common/wasm/wasm.h
```diff
@@ -844,6 +850,21 @@ inline bool Wasm::copyToPointerSize(const Buffer::Instance& buffer, uint64_t sta
...
```
PiotrSikora: Nit: you can remove this `ASSERT`, since `NOT_REACHED_GCOVR_EXCL_LINE` logs `"not reachable"` and calls `abort()`.

PR #132, opened 2019-08-15
Fix null plugin name

bianpengyuan:
This causes panic when load thread local instance from the base instance. See test failure here: https://github.com/istio/proxy/pull/2359 @jplevyak @PiotrSikora 

Do we have any test for null plugin?


bianpengyuan:
cc @mandarjog

PR #139, opened 2019-08-20
Add support for connectionInfo peerCertificatePresented().

jplevyak:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


jplevyak:
@mandarjog

api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -518,8 +516,16 @@ inline WasmResult getResponseTlsVersion(std::string *tls_version_ptr) {
   PROXY_EXPRESSION_GET_STRING("response.tls_version", tls_version_ptr);
 }
 
+inline WasmResult getRequestPeerCertificatePresented(bool *peer_certificate_presented) {
+  PROXY_EXPRESSION_GET("request.peer_certificate_presented", peer_certificate_presented);
+}
+
+inline WasmResult getResponsePeerCertificatePresented(bool *peer_certificate_presented) {
+  PROXY_EXPRESSION_GET("response.peer_certificate_presented", peer_certificate_presented);
+}
```
PiotrSikora: Erm, why do we have both: `"request.peer_certificate_presented"` and `"response.peer_certificate_presented"`?

mandarjog: There are 4 possibilities here
1. upstream.peer_certificate_present (which should always be true, if upstream did not present a certificate, tls connection will fail.  `== upstream.response.peer_certificate_presented`
2. upstream.my_cert_presented if this is true then client can say that the upstream connection is mtls.
3. downstream.peer_certificate_presented - server can assert than connection is mtls
4. downstream.my_cert_presented -- this should be always be true for mtls

@jplevyak which of these 4 are mapped onto the 2 you have implemented?

kyessenov: Upstream TLS info is not implemented. I'm working on it here https://github.com/envoyproxy/envoy/pull/7911

PiotrSikora: @mandarjog it's `local`, not `my`.

jplevyak: Both of these are "downstreamSslConnection()->peerCertifcatePresented coming from the Request and Response streamInfo respectively.   I would imaging that they mean the same thing since the connection.ssl() is going to be the same.  I exposed the Envoy API here directly, but it is clearly not what we want in the long run.  I am hoping that when Kuat replaces these "expressions" with the standard ones he can remove the incorrect ones.

PiotrSikora: Why are we exposing the same info in both request and response objects? Seems weird.

PiotrSikora:
@kyessenov is going to fix it all shortly, so I'll allow this duplication.

PR #142, opened 2019-08-20
add context traversal from WASM

kyessenov:

Second attempt at #115 

/assign @PiotrSikora 
/assign @jplevyak

### DIFF
--- a/api/wasm/cpp/proxy_wasm_api.h
+++ b/api/wasm/cpp/proxy_wasm_api.h
@@ -55,6 +55,7 @@ public:
   WasmData(const char* data, size_t size) : data_(data), size_(size) {}
   ~WasmData() { ::free(const_cast<char*>(data_)); }
   const char* data() { return data_; }
+  size_t size() { return size_; }
   StringView view() { return {data_, size_}; }
   std::string toString() { return std::string(view()); }
   int32_t int32() { return *reinterpret_cast<const int32_t*>(data_); }
@@ -520,6 +521,31 @@ inline WasmResult getResponseTlsVersion(std::string *tls_version_ptr) {
 
 inline WasmResult getPluginDirection(PluginDirection *direction_ptr) {
   PROXY_EXPRESSION_GET_UINT32("plugin.direction", reinterpret_cast<uint32_t*>(direction_ptr));
+}
+
+// Generic selector
+inline absl::optional<WasmDataPtr> getSelectorExpression(std::initializer_list<absl::string_view> parts) {
+  size_t size = 0;
+  for (auto part: parts) {
+    size += part.size() + 1; // null terminated string value
+  }
+
+  char* buffer = static_cast<char*>(::malloc(size));
+  char* b = buffer;
+
+  for (auto part : parts) {
+    memcpy(b, part.data(), part.size());
+    b += part.size();
+    *b++ = 0;
+  }
+
+  const char* value_ptr = nullptr;
+  size_t value_size = 0;
+  auto result = proxy_getSelectorExpression(buffer, size, &value_ptr, &value_size);
+  if (result != WasmResult::Ok) {
+    return {};
+  }
+  return std::make_unique<WasmData>(value_ptr, value_size);
 }
 
 // Metadata
--- a/api/wasm/cpp/proxy_wasm_api.h
+++ b/api/wasm/cpp/proxy_wasm_api.h
@@ -542,6 +542,7 @@ inline absl::optional<WasmDataPtr> getSelectorExpression(std::initializer_list<a
   const char* value_ptr = nullptr;
   size_t value_size = 0;
   auto result = proxy_getSelectorExpression(buffer, size, &value_ptr, &value_size);
+  ::free(buffer);
   if (result != WasmResult::Ok) {
     return {};
   }
--- a/source/extensions/common/wasm/null/sample_plugin/plugin.cc
+++ b/source/extensions/common/wasm/null/sample_plugin/plugin.cc
@@ -6,6 +6,7 @@
 #else
 
 #include "extensions/common/wasm/null/null_plugin.h"
+#include "absl/base/casts.h"
 
 namespace Envoy {
 namespace Extensions {
@@ -45,7 +46,24 @@ FilterDataStatus PluginContext::onRequestBody(size_t body_buffer_length, bool /*
 
 void PluginContext::onLog() {
   auto path = getRequestHeader(":path");
-  logWarn("onLog " + std::to_string(id()) + " " + std::string(path->view()));
+  if (path->view() == "/test_context") {
+    logWarn("request.path: " + getSelectorExpression({"request", "path"}).value()->toString());
+    logWarn("node.metadata: " +
+            getSelectorExpression({"node", "metadata", "istio.io/metadata"}).value()->toString());
+    logWarn("metadata: " + getSelectorExpression({"metadata", "filter_metadata", "envoy.wasm",
+                                                  "wasm_request_get_key"})
+                               .value()
+                               ->toString());
+    auto responseCode = getSelectorExpression({"response", "code"}).value();
+    if (responseCode->size() == sizeof(int64_t)) {
+      char buf[sizeof(int64_t)];
+      responseCode->view().copy(buf, sizeof(int64_t), 0);
+      int64_t code = absl::bit_cast<int64_t>(buf);
+      logWarn("response.code: " + absl::StrCat(code));
+    }
+  } else {
+    logWarn("onLog " + std::to_string(id()) + " " + std::string(path->view()));
+  }
 }
 
 void PluginContext::onDone() { logWarn("onDone " + std::to_string(id())); }
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -1094,8 +1094,6 @@ absl::optional<std::string> Context::getSelectorExpression(absl::string_view pat
   using Filters::Common::Expr::CelValue;
   using google::api::expr::runtime::FieldBackedListImpl;
   using google::api::expr::runtime::FieldBackedMapImpl;
-  using Protobuf::Descriptor;
-  using Protobuf::FieldDescriptor;
 
   bool first = true;
   CelValue value;
@@ -1161,8 +1159,8 @@ absl::optional<std::string> Context::getSelectorExpression(absl::string_view pat
       if (msg == nullptr) {
         return {};
       }
-      const Descriptor* desc = msg->GetDescriptor();
-      const FieldDescriptor* field_desc = desc->FindFieldByName(std::string(part));
+      const Protobuf::Descriptor* desc = msg->GetDescriptor();
+      const Protobuf::FieldDescriptor* field_desc = desc->FindFieldByName(std::string(part));
       if (field_desc == nullptr) {
         return {};
       } else if (field_desc->is_map()) {
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -1115,7 +1115,9 @@ absl::optional<std::string> Context::getSelectorExpression(absl::string_view pat
 
     size_t end = path.find('\0', start);
     if (end == absl::string_view::npos) {
-      end = path.size();
+      // this should not happen unless the input string is not null-terminated in the view
+      // if it does happen, assume malformed path and return empty
+      return {};
     }
     auto part = path.substr(start, end - start);
     start = end + 1;
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -475,9 +475,7 @@ Word getSelectorExpressionHandler(void* raw_context, Word path_ptr, Word path_si
   if (!path.has_value()) {
     return wasmResultToWord(WasmResult::InvalidMemoryAccess);
   }
-  std::vector<absl::string_view> parts =
-      absl::StrSplit(path.value(), absl::ByChar('\0'), absl::SkipEmpty());
-  auto value = context->getSelectorExpression(parts);
+  auto value = context->getSelectorExpression(path.value());
   if (value.has_value()) {
     context->wasm()->copyToPointerSize(value.value(), value_ptr_ptr, value_size_ptr);
     return wasmResultToWord(WasmResult::Ok);
@@ -1092,7 +1090,7 @@ absl::optional<std::string> serializeValue(Filters::Common::Expr::CelValue value
   return {};
 }
 
-absl::optional<std::string> Context::getSelectorExpression(std::vector<absl::string_view> parts) {
+absl::optional<std::string> Context::getSelectorExpression(absl::string_view path) {
   using Filters::Common::Expr::CelValue;
   using google::api::expr::runtime::FieldBackedListImpl;
   using google::api::expr::runtime::FieldBackedMapImpl;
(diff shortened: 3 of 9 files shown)


source/extensions/common/wasm/wasm.cc
```diff
@@ -462,6 +467,22 @@ Word setMetadataStructHandler(void* raw_context, Word type, Word name_ptr, Word
...
+Word getSelectorExpressionHandler(void* raw_context, Word path_ptr, Word path_size,
+                                  Word value_ptr_ptr, Word value_size_ptr) {
+  auto context = WASM_CONTEXT(raw_context);
+  auto path = context->wasmVm()->getMemory(path_ptr, path_size);
+  if (!path.has_value()) {
+    return wasmResultToWord(WasmResult::InvalidMemoryAccess);
+  }
+  auto value = context->getSelectorExpression(path.value());
+  if (value.has_value()) {
+    context->wasm()->copyToPointerSize(value.value(), value_ptr_ptr, value_size_ptr);
```
jplevyak: You need to handle the return value here and return a memory error on false.

source/extensions/common/wasm/wasm.cc
```diff
@@ -1016,6 +1037,153 @@ uint64_t Context::getCurrentTimeNanoseconds() {
...
+      response_trailers_ ? response_trailers_ : access_log_response_trailers_;
+
+  size_t start = 0;
+  while (true) {
+    if (start >= path.size()) {
+      break;
+    }
+
+    size_t end = path.find('\0', start);
+    if (end == absl::string_view::npos) {
```
jplevyak: This should never be hit since we always null-terminate.  I suggest returning an error here.

api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -522,6 +523,32 @@ inline WasmResult getPluginDirection(PluginDirection *direction_ptr) {
   PROXY_EXPRESSION_GET_UINT32("plugin.direction", reinterpret_cast<uint32_t*>(direction_ptr));
 }
 
+// Generic selector
+inline absl::optional<WasmDataPtr> getSelectorExpression(std::initializer_list<absl::string_view> parts) {
```
jplevyak: I thought we had decided on using the same representation that Pairs uses:
numberOfPathElements : uint32
pathElement0Size : uint32
pathElement1Size : uint32
...
pathElement0 : char[pathElement0Size]
pathElement1 : char[pathElement1Size]
...

This would be very fast to convert into a vector of string_view:

std::vector<string_view> path;
auto size = *reinterpret_cast<uint32_t*>(&buf[0]);
path.reserve(size);
for (int i = 0; i < size; i++) {
  path.push_back(string_view( ....
}

This encoding requires searching for the nulls, although that is not all that expensive since it can be done in 64-bit chunks for large path parts.

I am OK with this, it just isn't what I was expecting.

source/extensions/common/wasm/wasm.cc
```diff
@@ -1016,6 +1037,153 @@ uint64_t Context::getCurrentTimeNanoseconds() {
...
+    auto part = path.substr(start, end - start);
+    start = end + 1;
+
+    // top-level ident
+    if (first) {
+      first = false;
+      if (part == "metadata") {
+        value = CelValue::CreateMessage(&info.dynamicMetadata(), &arena);
+      } else if (part == "request") {
+        value = CelValue::CreateMap(Protobuf::Arena::Create<Filters::Common::Expr::RequestWrapper>(
```
jplevyak: Where is RequestWrapper defined?   Also isn't this very expensive: converting a structure into a proto just so we can use proto functions to access the values?

api/wasm/cpp/proxy_wasm_api.h
kyessenov: Yes, sorry. I was trying to make it allocation-free, and this is what it ended up with. I think it's nice to have a streaming-like access since it matches what the code does in the end. I had a vector at first, but then it looked like it wasn't necessary.

source/extensions/common/wasm/wasm.cc
kyessenov: It's in the base envoy. It's not a proto, it's a wrapper with no state.

source/extensions/common/wasm/wasm.cc
kyessenov: Thanks, done.

source/extensions/common/wasm/wasm.cc
```diff
@@ -1016,6 +1037,153 @@ uint64_t Context::getCurrentTimeNanoseconds() {
...
+    return std::string(out.buf, n);
+  }
+  case CelValue::Type::kUint64: {
+    const size_t n = sizeof(uint64_t);
+    marshall<n> out = absl::bit_cast<marshall<n>>(value.Uint64OrDie());
+    return std::string(out.buf, n);
+  }
+  case CelValue::Type::kDouble: {
+    const size_t n = sizeof(double);
+    marshall<n> out = absl::bit_cast<marshall<n>>(value.DoubleOrDie());
```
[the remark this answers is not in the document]
kyessenov: Hm, I'm confused. I thought `char[N]` is stack allocated where `N` is determined at compile time. I borrowed this trick from `absl`, but perhaps, I'm not reading it right.

source/extensions/common/wasm/wasm.cc
kyessenov: The code might look weird since I could not find a way to optionally stack allocate the wrapper.

source/extensions/common/wasm/wasm.cc
kyessenov: Changed to return immediately with empty result. Returning an error is rather painful since we don't have StatusOr.

api/wasm/cpp/proxy_wasm_api.h
jplevyak: you can still do it allocation free with the above data structure, but I agree that this isn't bad.  let's keep it

source/extensions/common/wasm/wasm.cc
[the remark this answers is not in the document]
jplevyak: You are right about that, but the implementation of absl::bit_cast uses a memcpy and then std::string does another.   You could avoid one of them by doing return std::string(reinterpret_cast<const char*>(&port), sizeof(port));  where port is the thing being converted into a string.

source/extensions/common/wasm/wasm.cc
jplevyak: https://en.cppreference.com/w/cpp/utility/variant (when we get to c++17).

You can return std::pair<WasmResult, std::string> or return the WasmResult and the std::string indirectly.

source/extensions/common/wasm/wasm.cc
jplevyak: Ah, I see, that code is from Aug 19th, no wonder I couldn't find it.

Just curious, what happens when you try to stack allocate it.  It seems to have a standard constructor.

kyessenov: Doesn't the body of the loop correspond to a code block? The compiler is free to delete the stack allocated value here, right? It might work, just worried whether that is introducing some undefined behavior.

jplevyak: I see what you are saying. You don't want to build the wrappers outside the loop because you aren't sure you are going to need them all.  This is a very reasonable solution.  The alternative of using alloca is not appealing in a loop.

source/extensions/common/wasm/wasm.cc
kyessenov: Done using output argument.

source/extensions/common/wasm/wasm.cc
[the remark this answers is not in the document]
kyessenov: Updated to avoid all mallocs.

jplevyak:
We are going to need an API next, we also need to replace all of the metadata code with this ABI.
Discussed with @PiotrSikora, let's remove the caching from the API level.  It could be a premature optimization.  Instead let us focus on coverage in preparation for upstreaming.  @mandarjog

source/extensions/common/wasm/null/sample_plugin/plugin.cc
```diff
@@ -45,7 +46,24 @@ FilterDataStatus PluginContext::onRequestBody(size_t body_buffer_length, bool /*
 
 void PluginContext::onLog() {
   auto path = getRequestHeader(":path");
-  logWarn("onLog " + std::to_string(id()) + " " + std::string(path->view()));
+  if (path->view() == "/test_context") {
```
PiotrSikora: I'm not a fan of mixing fake behavior required by tests in production code. Is there any reason why this cannot be created/configured/mocked in tests? I know this is a sample plugin, but people tend to copy&paste those.

At the very least, please make the path a bit more unique (`/test_<random sha>` would be acceptable).

kyessenov: I don't like this, too, but that's how the other tests do it. We need more mocks I think.

api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -522,6 +523,32 @@ inline WasmResult getPluginDirection(PluginDirection *direction_ptr) {
...
+    size += part.size() + 1; // null terminated string value
+  }
+
+  char* buffer = static_cast<char*>(::malloc(size));
+  char* b = buffer;
+
+  for (auto part : parts) {
+    memcpy(b, part.data(), part.size());
+    b += part.size();
+    *b++ = 0;
```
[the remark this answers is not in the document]
kyessenov: Yes, this encoding is only meant for symbols/strings. We should use a real marshalling protocol for binary data.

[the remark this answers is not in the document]
PiotrSikora: Ideally, we'd have one way for passing an array/list of any type of data.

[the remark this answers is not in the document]
kyessenov: Why? You're going to lose a lot of efficiency if you don't share prior knowledge about the type of data. Genericity is not free.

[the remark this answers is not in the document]
PiotrSikora: My point was that we could use serialization that works for both, that is, instead of using:
```
<string1.data>\0<string2.data>\0<string3.data>\0
Hello\0world\0!\0
```
we could be using:
```
<string1.len><string1.data><string2.len><string2.data><string3.len><string3.data>
\5Hello\5world\1!
```
which works for both strings and binary values.

I don't think you'd lose any efficiency in doing that (other than a few more bytes for the length).

But in order to unblock this, I'm fine with merging this as-is and moving the discussion to the ABI doc, once I ship it.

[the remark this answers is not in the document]
kyessenov: That's fair. Let's address this later. My concern is that it should support streaming/yielding. Other than that, any scheme is fine.

PiotrSikora:
@jplevyak any last minute comments? If not, please merge. Thanks!

PiotrSikora:
Erm, this broke the API:
```
test/extensions/wasm/test_data$ make
em++ -s WASM=1 -s BINARYEN_TRAP_MODE='clamp' -s LEGALIZE_JS_FFI=0 -s EMIT_EMSCRIPTEN_METADATA=1 --std=c++17 -O3 -g3 -DEMSCRIPTEN_PROTOBUF_LITE=1 -I/home/piotrsikora/Google/envoy/api/wasm/cpp -I/home/piotrsikora/Google/envoy/api/wasm/cpp/google/protobuf -I/usr/local/include --js-library /home/piotrsikora/Google/envoy/api/wasm/cpp/proxy_wasm_intrinsics.js logging_cpp.cc /home/piotrsikora/Google/envoy/api/wasm/cpp/proxy_wasm_intrinsics_lite.pb.cc /home/piotrsikora/Google/envoy/api/wasm/cpp/struct_lite.pb.cc  /home/piotrsikora/Google/envoy/api/wasm/cpp/libprotobuf-lite.bc -o logging_cpp.js
In file included from logging_cpp.cc:4:
In file included from /home/piotrsikora/Google/envoy/api/wasm/cpp/proxy_wasm_intrinsics.h:26:
/home/piotrsikora/Google/envoy/api/wasm/cpp/proxy_wasm_api.h:533:8: error: use of undeclared identifier 'absl'
inline absl::optional<WasmDataPtr> getSelectorExpression(std::initializer_list<absl::string_view> parts) {
       ^
/home/piotrsikora/Google/envoy/api/wasm/cpp/proxy_wasm_api.h:533:80: error: use of undeclared identifier 'absl'
inline absl::optional<WasmDataPtr> getSelectorExpression(std::initializer_list<absl::string_view> parts) {
                                                                               ^
/home/piotrsikora/Google/envoy/api/wasm/cpp/proxy_wasm_api.h:555:10: error: no viable conversion from returned value of type 'typename __unique_if<WasmData>::__unique_single' (aka 'unique_ptr<WasmData>') to function return type 'int'                    
  return std::make_unique<WasmData>(value_ptr, value_size);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
3 errors generated.
shared:ERROR: '/home/piotrsikora/Google/emsdk/fastcomp/fastcomp/bin/clang++ -target asmjs-unknown-emscripten -D__EMSCRIPTEN_major__=1 -D__EMSCRIPTEN_minor__=38 -D__EMSCRIPTEN_tiny__=41 -D_LIBCPP_ABI_VERSION=2 -Werror=implicit-function-declaration -Xclang -nostdsysteminc -Xclang -isystem/home/piotrsikora/Google/emsdk/fastcomp/emscripten/system/include/libcxx -Xclang -isystem/home/piotrsikora/Google/emsdk/fastcomp/emscripten/system/lib/libcxxabi/include -Xclang -isystem/home/piotrsikora/Google/emsdk/fastcomp/emscripten/system/include/compat -Xclang -isystem/home/piotrsikora/Google/emsdk/fastcomp/emscripten/system/include -Xclang -isystem/home/piotrsikora/Google/emsdk/fastcomp/emscripten/system/include/libc -Xclang -isystem/home/piotrsikora/Google/emsdk/fastcomp/emscripten/system/lib/libc/musl/arch/emscripten -Xclang -isystem/home/piotrsikora/Google/emsdk/fastcomp/emscripten/system/local/include -DEMSCRIPTEN --std=c++17 -O3 -DEMSCRIPTEN_PROTOBUF_LITE=1 -I/home/piotrsikora/Google/envoy/api/wasm/cpp -I/home/piotrsikora/Google/envoy/api/wasm/cpp/google/protobuf -I/usr/local/include -mllvm -disable-llvm-optzns logging_cpp.cc -Xclang -disable-O0-optnone -Xclang -isystem/home/piotrsikora/Google/emsdk/fastcomp/emscripten/system/include/SDL -c -o /tmp/emscripten_temp_zHxxPF/logging_cpp_0.o -emit-llvm' failed (1)
../../../../api/wasm/cpp/Makefile.base_lite:10: recipe for target 'logging_cpp.wasm' failed
make: *** [logging_cpp.wasm] Error 1
```

PiotrSikora:
Should be fixed by: https://github.com/envoyproxy/envoy-wasm/pull/149

kyessenov:
Can you document which part of the code is C++-17 and why absl is not allowed, please?
It's not at all clear from the style.

PiotrSikora:
Low-level API/SDK uses C++17, code in Envoy uses abseil.

source/extensions/common/wasm/wasm.cc
```diff
@@ -1016,6 +1040,153 @@ uint64_t Context::getCurrentTimeNanoseconds() {
...
 }
 
+WasmResult serializeValue(Filters::Common::Expr::CelValue value, std::string* result) {
+  using Filters::Common::Expr::CelValue;
+  switch (value.type()) {
+  case CelValue::Type::kMessage:
+    if (value.MessageOrDie() != nullptr) {
+      value.MessageOrDie()->SerializeToString(result);
+      return WasmResult::Ok;
+    }
```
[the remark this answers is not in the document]
kyessenov: Not intentional. Thanks for fixing.

PR #144, opened 2019-08-21
Add LOG macros with line number and file name.

jplevyak:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


jplevyak:
@mandarjog

api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -49,6 +49,15 @@ inline void logAbort(StringView logMessag) {
   abort();
 }
 
+#define LOG(_level, ...)                                                        \
+    log##_level(std::string("[") + __FILE__ + ":" + std::to_string(__LINE__) + "]::" + __FUNCTION__ + "() " + __VA_ARGS__)
```
PiotrSikora: For the reference, this is logged as:
```
[2019-08-24 00:12:08.944][29379][trace][wasm] [source/extensions/common/wasm/wasm.cc:1718] wasm log my_vm_id my_root_id: [envoy_filter_http_wasm_example.cc:30]::onStart() onStart
```
I think we could use `[]` around identifiers and probably drop the `::`.

Other than that, it looks fine, so feel free to merge if you disagree.

For the record, I might need to bike-shed this shortly anyway, to align it with host-environment logging in V8.

mandarjog: Already better than just wasm.cc:1718 .
Would be even better to elide wasm.cc since is occupies so much real estate .

jplevyak: I'll merge this and we can bikeshed later.  Still some work to be done here.

PR #149, opened 2019-08-24
Fix after previous commit.

PiotrSikora:


PiotrSikora:
> It's probably a good idea to add a linter to ban absl from low-level ABI.

Agreed. Patches welcomed!

With a note that we're going to move the API/SDK to the separate repo before upstreaming anyway. cc @jplevyak

PR #152, opened 2019-08-25
use foreign_cc to build wee8

lizan:

Description:
To propagate CC context better down to gn/ninja.

Additionally, added use_glib=false so glib is no longer needed.

Risk Level:
Testing:
Docs Changes:
Release Notes:


PiotrSikora:
Did you see #151? It should address all `CC`-related problems that you might have.

This is a nice PR, I actually didn't know you can override `configure_make` this way, so thanks for doing that! Having said that, I'd rather keep this hermetic and working as tested, for the time being, since:
- This is going to be migrated to BUILD files soon, anyway (I have a working version, but I'm still trying to figure out a weird ASan error).
- I have a few more PRs staged (pending #151) to hook sanitizers, update V8, etc. that I'm not sure will work with this approach,
- I'm trying to get this shipped in 1.3 (i.e. over the next few days), so I'd prefer to avoid unnecessary changes right now,
- I'm not sure if this solves anything that isn't already addressed by #151, since it seems that the `make` step is still executed as a single job and not distributed.

But I'm happy to discuss if you disagree.

PiotrSikora:
Since this is separate from this discussion, could you send a separate PR with `use_glib=false` (and revert CircleCI image to `envoyproxy/envoy-build:cd8574de783791b3353579b489222bfda74888da`)?

PiotrSikora:
Also, this breaks with GCC (see #151 for how it's fixed there with `is_clang`).

lizan:
I tested with libc++ and this work as is. Didn’t test gcc but I guess a couple of select should make it work.

Sanitizer flags should be already passed down in this way so I guess ASAN is already working from the CI result, though I haven’t verified it.

bazel/foreign_cc/wee8.patch
```diff
@@ -3,8 +3,9 @@
 # 3. Fix for VMs with overlapping lifetimes (https://crrev.com/c/1698387).
 # 4. Fix linking with unbundled toolchain on macOS.
 # 5. Increase VSZ limit to 4TiB (allows us to start up to 370 VMs).
---- a/wee8/build/toolchain/gcc_toolchain.gni
-+++ b/wee8/build/toolchain/gcc_toolchain.gni
+# 6. Remove Bazel added -fuse-ld=gold if gn chose -fuse-ld=lld
```
PiotrSikora: Please add: `# 7. Disable :compiler_deterministic, because ...` (or `6.`, really).

bazel/foreign_cc/wee8.patch
```diff
@@ -92,3 +93,28 @@
    void make_data();
    void free_data();
  #else
+
+
```
PiotrSikora: Nit: extra empty lines between patches.

bazel/foreign_cc/BUILD
```diff
@@ -290,3 +290,22 @@ envoy_cmake_external(
         "//conditions:default": ["libz.a"],
     }),
 )
+
+configure_make(
+    name = "wee8",
```
PiotrSikora: Nit: sort alphabetically.

PiotrSikora:
@lizan do you want to finish this before I start upstreaming V8, or should I upstream it as-is and migrate to `BUILD` files in a few weeks?

lizan:
@PiotrSikora let's drop this and migrating to native BUILD that you have :)

PR #153, opened 2019-08-25
build: drop GLib2 dependency in wee8.

PiotrSikora:
Suggested by Lizan in:
https://github.com/envoyproxy/envoy-wasm/pull/152


PiotrSikora:
Ah, I forgot about WAVM, which still requires newer CMake...

PR #171, opened 2019-08-30
wasm: add support for network filters (read-only).

PiotrSikora:


PiotrSikora:
@jplevyak this is identical to #170, but with Envoy interfaces implemented in `NetworkFilters::Wasm::Filter` instead of in `Wasm::Context` (see the last commit).

jplevyak:
I am not sure what I think of this.  It adds another layer, but doesn't end up appreciably simplifying Wasm::Context.  It makes calls into the WASM module and mocking calls out slightly more difficult, because now the events are injected either on a different object with the standard interface (e.g. onData here) or on the Context object with a different interface (onDownstreamData).  In either case we need to have to mock the Wasm::Context.  It also isn't clear that we need both onData and onDownstreamData.  If we eliminate the latter then not adding the layer would be simpler overall.

In any case if we make this change we should make it over all the filter types.

Let's table this PR for now and open an issue to discuss this.  I am also not sure which way the upstream reviewers will go.

I am going to add another layer to differentiate which configured filter is responsible for a particular Wasm::Context.  Let's look at that before we decide on this since I think they are related.

PiotrSikora:
Abandoned in favor of #170, see above.

PR #172, opened 2019-08-30
Add simple wasm performance test.

jplevyak:

Optimized results:

Benchmark                                                                Time             CPU       Iterations 
BM_WasmSimpleCallSpeedTest/V8SpeedTest           273 ns          273 ns       2591211
BM_WasmSimpleCallSpeedTest/WavmSpeedTest     221 ns           221 ns      3144155
BM_WasmSimpleCallSpeedTest/NullSpeedTest         107 ns          107 ns       6260650

Net: the cost of calls into and out of the VMs are very small.     WAVM is faster than V8.


AndrewScheidecker:
Based on some investigation on WAVM's `invoke-bench`, I think `catchRuntimeExceptions` is probably slower than `invokeFunctionUnchecked` in your benchmark. It calls `unwindSignalsAsExceptions`, which calls `Platform::catchSignals`, which was quite slow: I've just made [a commit](https://github.com/WAVM/WAVM/commit/c5e1ee0153651eb5df7c38332691548d9d2f4473) that makes it 10x faster.

FYI you're using a version of the WAVM API from before [a change](https://github.com/WAVM/WAVM/commit/fd3ed14a3ad406080dd3b75bb37752b8274b4ed0) I made recently to merge `invokeFunctionChecked` and `invokeFunctionUnchecked`.

The unified `invokeFunction` is slightly slower than `invokeFunctionUnchecked`, mostly because it calls `unwindSignalsAsExceptions` to ensure that if you have any C++ objects on the stack between the `invokeFunction` and `catchRuntimeExceptions`, that their destructor will be called. The combined overhead of `catchRuntimeExceptions` and `invokeFunction` should be much lower with my `catchSignals` optimization, though.

jplevyak:
I tried to update our WAVM implementation, but the new version is aborting and/or corrupting memory.  In particular I can build a test case where  std::to_string(7) returns "3454" (or some other bogus result).  I changed the code to use char buf[16]; sprintf(buf, "%d", 7); and it also gets corrupted.  This doesn't happen in V8 and it doesn't happen in the older version of WAVM. 

If you  are curious: https://github.com/jplevyak/envoy-wasm/pull/new/wasm-wavmupgrade

The test //test/extensions/wasm:wasm_test fails.

Any suggestions on how to debug would be appreciated.

@AndrewScheidecker

AndrewScheidecker:
> I tried to update our WAVM implementation, but the new version is aborting and/or corrupting memory. In particular I can build a test case where std::to_string(7) returns "3454" (or some other bogus result). I changed the code to use char buf[16]; sprintf(buf, "%d", 7); and it also gets corrupted. This doesn't happen in V8 and it doesn't happen in the older version of WAVM.

To be clear, this is for `std::to_string`/`sprintf` in some WASM code hosted by Envoy/WAVM?

Can you reproduce it with a WASM file that works with plain WAVM?

Does it still happen if you use the WAVM commit immediately before this one? https://github.com/WAVM/WAVM/commit/b9d2a112cfd8e9404285dd13f4f1849539bcb5f3

jplevyak:
Problem solved.  I was not initializing the heap pointer correctly for emscripten.

jplevyak:
@PiotrSikora ready for review.

PiotrSikora:
@jplevyak could you merge with master and re-run tests to get accurate numbers as of now?

test/extensions/wasm/wasm_speed_test.cc
```diff
@@ -0,0 +1,95 @@
+/**
+ * Simple WASM speed test.
+ *
+ * Run with:
+ * bazel run -c opt  --linkopt -fuse-ld=lld --define wasm=enabled --config=libc++ wasm_speed_test
```
PiotrSikora: This should be really be:
```
TEST_SRCDIR=`pwd` TEST_WORKSPACE=bazel-$(basename `pwd`) bazel run --define wasm=enabled --config=libc++ -c opt //test/extensions/wasm:wasm_speed_test
```
1. You should really use the full path when describing Bazel targets.
2. `--linkopt -fuse-ld=lld` shouldn't be necessary (at least not universally).
3. `TEST_SRCDIR` and `TEST_WORKSPACE` are needed to run this.

jplevyak: --linkopt -fuse-ld=lld is required on my box

If I include your definition of TEST_SRCDIR it can't find speed_cpp.wasm. I got the code for TEST_RUNDIR from envoy/test/main.cc.  What works for me is setting TEST_SRCDIR to the base envoy directory (i.e. where it was cloned).

I don't see how TEST_WORKSPACE can be "bazel-wasm" (TEST_WORKSPACE=bazel-$(basename `pwd`))  that directory doesn't appear anywhere in the workspace or in .cache/bazel

jplevyak: Well, I stand corrected.  You solution works when I run from the root of the tree.

jplevyak: Updated to the latest master.  Updated the run comments.  Left a note that lld might be required.

PTAL

test/extensions/wasm/wasm_speed_test.cc
```diff
@@ -0,0 +1,97 @@
+/**
+ * Simple WASM speed test.
+ *
+ * Run with:
+ * TEST_SRCDIR=`pwd` TEST_WORKSPACE=bazel-$(basename `pwd`) bazel run --define wasm=enabled
+ * --config=libc++ -c opt //test/extensions/wasm:wasm_speed_test
+ * Note: "--linkopt -f--fuse-ld" may be required as well depending on the build environment.
```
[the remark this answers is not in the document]
jplevyak: Done.

jplevyak:
Upgraded the version of protobuf which was used with speed_cpp.wasm.

PR #177, opened 2019-09-04
Add toolchain support for WASM BUILD files.

jplevyak:
Works up to linking.

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


jplevyak:
Any suggestions @kyessenov ?  I got this up to the point where

cd envoy/examples/wasm
bazel build --config=wasm envoy_filer-wasm_http_example.wasm

returns a link error.  There are a couple problems that I am having:

1) setting EM_CACHE=tmp/emscripten cache results in shutil.copyfile not knowing that the src and dest are the same, so that it throws an exception when trying to cache struct_info.json.  I worked around this by using the absolute path /tmp/emscripten_cache

2) it doesn't link because we need to have a proxy_wasm_intrinsics.js available for linking (it describes the external functions), but bazel really doesn't want to allow .js files to compile/link dependencies.  

I also need to add the protobuf .bc files as linked dependencies.

Perhaps I just need to make a wasm_cc_binary target. 

Suggestions?

jplevyak:
I am going to drop this.  We are going to move the SDK to another repo instead.

PR #179, opened 2019-09-04
use filter state for storing shared filter state

kyessenov:
- use filter state wrapper around a generic value to avoid the implicit proto merge inside `setDynamicMetadata`.
- replace `PluginDirection` with the upstream `TrafficDirection`
- combine metadata selector expression with the get selector handler

### DIFF
--- a/test/extensions/filters/http/wasm/test_data/metadata_cpp.cc
+++ b/test/extensions/filters/http/wasm/test_data/metadata_cpp.cc
@@ -34,7 +34,7 @@ FilterHeadersStatus ExampleContext::onRequestHeaders() {
   if (r != WasmResult::Ok) {
     logDebug(toString(r));
   }
-  r = setMetadataStringValue(MetadataType::Request, "wasm_request_set_key", "wasm_request_set_value");
+  r = setMetadataStringValue("wasm_request_set_key", "wasm_request_set_value");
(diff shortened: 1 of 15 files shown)


kyessenov:
Fixed tests and recompiled the test modules. Should be ready for review now.

kyessenov:
cc @mandarjog 

The shared state uses envoy's filter state, which is a map string -> object. For object we use a dummy wrapper around protobuf value, and the key is the same parameter key.

jplevyak:
This look find to me as long as @mandarjog is good with it.

test/extensions/filters/http/wasm/test_data/metadata_cpp.cc
```diff
@@ -34,7 +34,7 @@ FilterHeadersStatus ExampleContext::onRequestHeaders() {
   if (r != WasmResult::Ok) {
     logDebug(toString(r));
   }
-  r = setMetadataStringValue(MetadataType::Request, "wasm_request_set_key", "wasm_request_set_value");
+  r = setStateStringValue("wasm_request_set_key", "wasm_request_set_value");
```
mandarjog: nit: setStateStringValue is a difficult name to parse. What is the lifetime of this "set". Is it state of the request or the filter?

Issue #183, opened 2019-09-11
Add a validateConfig() API


mandarjog:
A wasm extension implements `onConfigure(std::unique_ptr<WasmData> configuration)` 
If configuration fails, there is no indication that something has gone wrong except a WARN message from the extension itself.

1.  `onConfigure()` should return a bool
2.  Add a `validateConfig(std::unique_ptr<WasmData> configuration)` API which is similar to `onConfigure` 
    *  It must do structural validation of the input configuration.
    *  It must be stateless, must be a pure function and must not use the HOST api.

A WASM extension that supports validateConfig can be used outside the proxy as a config validator.

mandarjog:
@kyessenov @douglas-reid

PR #190, opened 2019-09-13
Add a bool return from onConfigure and add onValidateConfiguration().

jplevyak:
https://github.com/envoyproxy/envoy-wasm/issues/183

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


jplevyak:
@mandarjog

jplevyak:
FYI: as a sideeffect of me improving my dev environment to auto-format files, the API file is now being formatted in Envoy style.  If this makes the review too difficult tell me and I'll do that separately.

jplevyak:
Ping

source/extensions/common/wasm/null/null_plugin.cc
```diff
@@ -219,7 +207,19 @@ void NullPlugin::getFunction(absl::string_view function_name, WasmCallWord<1>* f
 }
 
 void NullPlugin::getFunction(absl::string_view function_name, WasmCallWord<3>* f) {
-  if (function_name == "_proxy_onRequestBody") {
+  if (function_name == "_proxy_onValidateConfiguration") {
+    auto plugin = this;
+    *f = [plugin](Common::Wasm::Context* context, Word context_id, Word ptr, Word size) {
+      SaveRestoreContext saved_context(context);
+      return Word(plugin->onConfigure(context_id.u64_, ptr.u64_, size.u64_));
```
PiotrSikora: Is this intentional?

source/extensions/common/wasm/wasm.h
```diff
@@ -189,8 +189,9 @@ class Context : public Http::StreamFilter,
   //
   // VM level downcalls into the WASM code on Context(id == 0).
   //
+  virtual bool onValidateConfiguration(absl::string_view configuration);
   virtual void onStart(absl::string_view root_id, absl::string_view vm_configuration);
-  virtual void onConfigure(absl::string_view configuration);
+  virtual bool onConfigure(absl::string_view configuration);
```
PiotrSikora: Shouldn't this be `WasmResult`?

source/extensions/common/wasm/wasm.cc
```diff
@@ -2180,6 +2185,7 @@ void Wasm::getFunctions() {
 #undef _GET
 
 #define _GET_PROXY(_fn) wasm_vm_->getFunction("_proxy_" #_fn, &_fn##_);
+  _GET_PROXY(onValidateConfiguration);
```
PiotrSikora: Since this is expected to be used in control planes, I'm not sure if there is any reason to plug this all the way inside Envoy (i.e. it should probably be limited to Proxy-WASM SDK, but not host implementation).

api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -315,8 +315,14 @@ class RootContext : public ContextBase {
   RootContext* asRoot() override { return this; }
   Context* asContext() override { return nullptr; }
 
+  // Can be used to validate the configuration (e.g. in the control plane). Returns false if the
+  // configuration is invalid.
+  virtual bool onValidateConfiguration(std::unique_ptr<WasmData> /* configuration */) {
```
PiotrSikora: Nit: not sure if `on` prefix is justified here.

source/extensions/common/wasm/null/null_plugin.cc
jplevyak: Fixed.

source/extensions/common/wasm/wasm.h
jplevyak: These are calls into the VM.  WasmResult is for calls coming out of the VM current (environment error codes).  Do we want to conflate the two?  Do we want a different enum or is a bool OK?

api/wasm/cpp/proxy_wasm_api.h
jplevyak: changed to validateConfiguration

source/extensions/common/wasm/wasm.cc
jplevyak: It doesn't hurt to have it here, so perhaps we can leave it until we figure out how this is going to be used.  @mandarjog

test/extensions/wasm/test_data/bad_signature_cpp.cc
```diff
@@ -3,9 +3,10 @@
 
 #include "proxy_wasm_intrinsics.h"
 
-extern "C" EMSCRIPTEN_KEEPALIVE void proxy_onConfigure(uint32_t, int bad, char* configuration,
-                                                       int size) {
+extern "C" EMSCRIPTEN_KEEPALIVE uint32_t proxy_onConfigure(uint32_t, int bad, char* configuration,
```
PiotrSikora: Nit: you could have left it as `void proxy_onConfigure(uint32_t, char* configuration, int size)`, since it's supposed to be bad signature anyway.

source/extensions/common/wasm/wasm.h
PiotrSikora: Good point.

test/extensions/wasm/test_data/bad_signature_cpp.cc
jplevyak: Yes, but does it reverts the approval if I change anything... ?

PiotrSikora: Probably.

PR #193, opened 2019-09-16
serialize maps and lists

kyessenov:

Follow up to resolver ABI change. Add the missing feature to return the entire struct now that it's not part of metadata.

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


kyessenov:
cc @bianpengyuan 

Please review. It's blocking the update to istio/proxy to use filter state (it turns out CEL converts structs to dynamic maps on the fly, so we started exercising a new path).

source/extensions/common/wasm/wasm.cc
```diff
@@ -1024,6 +1024,77 @@ uint64_t Context::getCurrentTimeNanoseconds() {
       .count();
 }
 
+// TODO(kyessenov) this is available but not exposed in the upstream, need to correct this
```
jplevyak: Do we have an issue for this?  If not let's create one and reference it here.

source/extensions/common/wasm/wasm.cc
```diff
@@ -1068,11 +1139,20 @@ WasmResult serializeValue(Filters::Common::Expr::CelValue value, std::string* re
     result->assign(reinterpret_cast<const char*>(&out), sizeof(absl::Time));
     return WasmResult::Ok;
   }
+  case CelValue::Type::kMap:
+  case CelValue::Type::kList: {
+    ProtobufWkt::Value out;
+    if (!exportValue(value, &out)) {
+      return WasmResult::SerializationFailure;
```
jplevyak: Is this the correct error?  What causes exportValue to fail?

source/extensions/common/wasm/wasm.cc
kyessenov: Done.

source/extensions/common/wasm/wasm.cc
kyessenov: I think so. For now, it's enums and errors. I am not sure what the right export value for errors should be.

kyessenov:
Can it be merged? I don't write access.

source/extensions/common/wasm/wasm.cc
```diff
@@ -1026,6 +1026,77 @@ uint64_t Context::getCurrentTimeNanoseconds() {
...
+      if (!exportValue((*map)[(*list)[i]].value(), &field_value)) {
+        return false;
+      }
+      (*struct_obj->mutable_fields())[field_key.string_value()] = field_value;
+    }
+    return true;
+  }
+  default:
+    // do nothing for special values
+    break;
```
PiotrSikora: Nit: `return false` (I know it's going to fall-through, but it's more readable and follows style in `serializeValue`).

kyessenov: can I do that later? i want to un-break istio/proxy since there is an ABI change in envoy-wasm that wasn't propagated.

PiotrSikora: Sure.

PR #197, opened 2019-09-18
Remove metadata stubs

### LINKED ISSUE #156: Remove old Metadata code.
We should migrate away from the old Metadata code, now that #142 has been merged.

kyessenov:
Fixes #156


kyessenov:
Red makes me happy.

api/wasm/cpp/proxy_wasm_externs.h
```diff
@@ -60,13 +60,7 @@ extern "C" WasmResult proxy_setTickPeriodMilliseconds(uint32_t millisecond);
...
 
-// Metadata
-extern "C" WasmResult proxy_getMetadata(MetadataType type, const char* key_ptr, size_t key_size,
-                                      const char** value_ptr_ptr, size_t* value_size_ptr);
-extern "C" WasmResult proxy_getMetadataPairs(MetadataType type, const char** value_ptr,
-                                           size_t* value_size);
-extern "C" WasmResult proxy_getMetadataStruct(MetadataType type, const char* name_ptr, size_t name_size,
-                                            const char** value_ptr_ptr, size_t* value_size_ptr);
+// Shared state
 extern "C" WasmResult proxy_setState(const char* key_ptr, size_t key_size,
```
jplevyak: What are we going to do about setState ?  AFAICT we need to be able to set bits of the state, like we can get bits with getSelectorExpression.  Why don't we have a setSelectorExpression ?

kyessenov: I think it can be done. I think the main question is, what happens if setSelectorExpression(filter_state, key, field_key) and field_key is already populated with a struct in key. Do we do proto merge? Or override the key?

kyessenov: @mandarjog for thoughts on that. I can do a follow-up to change setState into something more generic.

jplevyak: Great.  Open an issue and let's get this in.

kyessenov: Filed #198

kyessenov:
Ping.

kyessenov:
Renamed the ABI handlers (not implementing full setProperty for now).

PR #204, opened 2019-09-24
Sync with master (c5738e1f85).

PiotrSikora:
Redo #195, which was accidentally squashed and makes merging with upstream really painful.


PiotrSikora:
Merging this directly to avoid squashing (there are no changes here anyway, only history rewrite).

cc @jplevyak

PR #208, opened 2019-09-25
Add a WasmPlugin layer for filter/service specific configuration.

jplevyak:


source/extensions/common/wasm/wasm.cc
```diff
@@ -2127,20 +2128,35 @@ void Wasm::getFunctions() {
...
 
-bool Wasm::initialize(const std::string& code, absl::string_view name, bool allow_precompiled) {
+bool Wasm::initialize(const std::string& code, bool allow_precompiled) {
   if (!wasm_vm_) {
     return false;
   }
+
+  // If the configured_vm_id is empty, then hash the code to create a unique vm_id.
+  if (vm_id_.empty()) {
+    std::vector<uint8_t> digest(SHA256_DIGEST_LENGTH);
```
PiotrSikora: Please create a variant of `Envoy::Common::Crypto::Utility::getSha256Digest()` that takes `const std::string&` or `absl::string_view` as an argument and use it here instead of an inline implementation.

api/envoy/config/accesslog/v2/wasm.proto
```diff
@@ -16,13 +16,5 @@ import "envoy/config/wasm/v2/wasm.proto";
...
-  // A unique ID for a set of filters/services in a VM which will share Context(s).
-  string root_id = 1;
-  // A unique ID for a VM so that multiple filters/services can call into the same VM.
-  string vm_id = 2;
-  //  Wasm configuration string e.g. a serialized protobuf which will be the argument to the
-  // VM configure() call.
-  string configuration = 3;
-  // Configuration for starting a new VM to be associated with the given vm_id.
-  envoy.config.wasm.v2.VmConfig vm_config = 4;
+  envoy.config.wasm.v2.PluginConfig config = 1;
```
PiotrSikora: Note: Envoy uses `extensions` (vs `modules` in Apache/IIS/lighttpd/LiteSpeed/NGINX/Varnish and `plugins` in ATS), so I'm not sure if we should be using `plugin` here... but let's leave this discussion for upstreaming.

jplevyak: Sounds good to me.

source/extensions/common/wasm/wasm.cc

jplevyak:
I think I have answered the comments.... PTAL @PiotrSikora

api/envoy/config/wasm/v2/wasm.proto
```diff
@@ -10,43 +10,60 @@ import "envoy/api/v2/core/base.proto";
...
+  // A unique ID for a VM permitting multiple filters/services can share the same VM.
+  // If left blank the 'vm_id' will be a hash of the Wasm code enabling the VM to persist over
+  // reconfigurations if the code does not change. This will also result in filters/services running
+  // the same Wasm code sharing the same VM. If set, 'vm_id' should include a version number
+  // otherwise the VM will not be reloaded if the Wasm code changes.
+  string vm_id = 1;
+
   // The Wasm VM type (see source/extensions/commmon/wasm/well_known_names.h).
-  string vm = 1;
+  string vm = 2;
```
PiotrSikora: Since we're already changing protos, maybe rename this to `runtime` (or `engine`)?

source/extensions/common/wasm/wasm.cc
```diff
@@ -2159,20 +2174,26 @@ void Wasm::getFunctions() {
...
   getFunctions();
 }
 
-bool Wasm::initialize(const std::string& code, absl::string_view name, bool allow_precompiled) {
+bool Wasm::initialize(const std::string& code, bool allow_precompiled) {
   if (!wasm_vm_) {
     return false;
   }
+
+  // If the configured_vm_id is empty, then hash the code to create a unique vm_id.
```
PiotrSikora: Nit: `configured_vm_id_` (missing `_`).

But also, it seems that the comment should say `vm_id_` and not `configured_vm_id_`.

source/extensions/common/wasm/wasm.cc
```diff
@@ -2595,105 +2617,96 @@ void GrpcStreamClientHandler::onRemoteClose(Grpc::Status::GrpcStatus status,
...
-    Upstream::ClusterManager& cluster_manager, Event::Dispatcher& dispatcher, Api::Api& api,
-    Stats::Scope& scope, envoy::api::v2::core::TrafficDirection direction,
-    const LocalInfo::LocalInfo& local_info, const envoy::api::v2::core::Metadata* listener_metadata,
-    Stats::ScopeSharedPtr scope_ptr, std::unique_ptr<Context> root_context_for_testing) {
-  auto wasm = std::make_shared<Wasm>(vm_config.vm(), vm_id, vm_config.configuration(),
-                                     cluster_manager, dispatcher, scope, direction, local_info,
-                                     listener_metadata, scope_ptr);
+static std::shared_ptr<Wasm> createWasmInternal(const envoy::config::wasm::v2::VmConfig& vm_config,
+                                                PluginSharedPtr plugin,
+
```
PiotrSikora: Nit: empty line.

source/extensions/common/wasm/wasm.h
```diff
@@ -427,21 +453,19 @@ class Wasm : public Envoy::Server::Wasm,
...
-  bool initialize(const std::string& code, absl::string_view name, bool allow_precompiled);
+  bool initialize(const std::string& code, bool allow_precompiled = false);
   bool configure(Context* root_context, absl::string_view configuration);
-  Context* start(absl::string_view root_id,
-                 absl::string_view vm_configuration); // returns the root Context.
+  Context* start(); // returns the root Context.
 
-  absl::string_view id() const { return id_; }
+  absl::string_view vm_id() const { return vm_id_; }
+  absl::string_view configured_vm_id() const { return configured_vm_id_; }
```
PiotrSikora: What's the point of `configured_vm_id`? I don't see it being used anywhere, and it's always the same value as `vm_id`.

api/envoy/config/wasm/v2/wasm.proto
jplevyak: Done: runtime

source/extensions/common/wasm/wasm.cc
jplevyak: Removed

source/extensions/common/wasm/wasm.h
jplevyak: Done

source/extensions/common/wasm/wasm.cc

PiotrSikora:
What happened to the hash in case `vm_id` is not configured?

PR #210, opened 2019-09-26
Add scripts for building all the tests and examples using the docker SDK.

jplevyak:


examples/wasm/docker_build_cpp_wasm.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+docker run -e uid="$(id -u)" -e gid="$(id -g)" -v $PWD:/work -v $(realpath ../../api/wasm/cpp):/external_sdk -w /work wasmsdk:v1 bash ./docker_cpp_builder.sh
```
PiotrSikora: Would something like `-u "$(id -u):$(id -g)"` work, so that we can get rid of the `chown` hack?

jplevyak: No, the problem is that the /root/emsdk/emsdk_env.sh doens't work for other users even if you chmod it to be readable for all.  It looks like it sets the paths relative to $HOME for the user calling it.  It is likely possible to hack it up somehow, but given that we are going to move to a separate repo with bazel I didn't want to spend the effort.

PiotrSikora: ACK.

jplevyak:
Done.

Issue #213, opened 2019-09-27
WAVM & Emscripten 1.38.46 regression


PiotrSikora:
`WasmTest.StatsHighLevel` fails under WAVM, but works fine under V8, when `stats_cpp.wasm` is compiled with Emscripten 1.38.46 (both work fine with 1.38.45).

```
[ RUN      ] Runtimes/WasmTest.StatsHighLevel/1
get counter = 1
get counter = 2
get counter = 5
get gauge = 2
Caught exception "emscripten abort" in WASM
test/extensions/wasm/wasm_test.cc:318: Failure
Actual function call count doesn't match EXPECT_CALL(*context, scriptLog_(spdlog::level::err, Eq("h_id = int_tag.7.string_tag.test_tag.bool_tag.true.test_histogram")))...
         Expected: to be called once
           Actual: never called - unsatisfied and active
Stack trace:
  0x35bde19: testing::internal::GoogleTestFailureReporter::ReportFailure()
  0x1de3458: testing::internal::Expect()
  0x35c434d: testing::internal::UntypedFunctionMockerBase::VerifyAndClearExpectationsLocked()
  0x49e6a3: testing::internal::FunctionMocker<>::~FunctionMocker()
  0x4a70d8: Envoy::Extensions::Wasm::TestContext::~TestContext()
  0x49e740: Envoy::Extensions::Wasm::TestContext::~TestContext()
  0x49e789: Envoy::Extensions::Wasm::TestContext::~TestContext()
  0x4a788f: std::__1::default_delete<>::operator()()
  0x4a780f: std::__1::unique_ptr<>::reset()
  0x45d529: std::__1::unique_ptr<>::~unique_ptr()
  0x45b70b: Envoy::Extensions::Wasm::WasmTest_StatsHighLevel_Test::TestBody()
  0x35b592e: testing::internal::HandleSehExceptionsInMethodIfSupported<>()
  0x35a4a7b: testing::internal::HandleExceptionsInMethodIfSupported<>()
  0x3591a63: testing::Test::Run()
  0x3592442: testing::TestInfo::Run()
... Google Test internal frames ...

test/extensions/wasm/wasm_test.cc:319: Failure
Actual function call count doesn't match EXPECT_CALL(*context, scriptLog_(spdlog::level::err, Eq("stack_c = 1")))...
         Expected: to be called once
           Actual: never called - unsatisfied and active
Stack trace:
[...]

PR #215, opened 2019-09-27
Update protobuf to v3.9.1.

PiotrSikora:
Built using:

   git clone https://github.com/protocolbuffers/protobuf
   cd protobuf
   git checkout v3.9.1
   git clone https://github.com/kwonoj/protobuf-wasm patches
   git apply patches/*.patch
   sh autogen.sh
   emconfigure ./configure CXXFLAGS="-O3"
   emmake make

with Emscripten 1.38.46 (fastcomp).


PiotrSikora:
@jplevyak since `.wasm` files are built using 1.38.46, 1 test is broken under WAVM (see: #213), but we don't exercise tests under WAVM on the CI... Let me know if you prefer that I regenerate everything using 1.38.45, where everything works.

jplevyak:
Yes, please regerate so that WAVM is not broken.  I am trying to find the problem.

jplevyak:
Did the image work?  I am concerned that the protobuf patch doesn't work cleanly (I have had problems with it before).

PiotrSikora:
Regenerated everything with Emscripten 1.38.45 (fastcomp).

PiotrSikora:
> Yes, please regerate so that WAVM is not broken. I am trying to find the problem.

Done.

> Did the image work? I am concerned that the protobuf patch doesn't work cleanly (I have had problems with it before).

I didn't use any image. I only updated the Emscripten version number there.

jplevyak:
Could you test the resulting image?

PiotrSikora:
Switched back to 1.38.46, since all tests pass with WAVM now.

The Docker image built fine, and I verified that it can regenerate `.wasm` files using scripts provided in #210.

PR #220, opened 2019-09-28
Upgrade WAVM.

jplevyak:


source/extensions/common/wasm/wasm.cc
```diff
@@ -2110,7 +2110,7 @@ void Wasm::establishEnvironment() {
 
     global_table_base_ = wasm_vm_->makeGlobal("env", "__table_base", Word(0));
     global_dynamictop_ =
-        wasm_vm_->makeGlobal("env", "DYNAMICTOP_PTR", Word(emscripten_dynamictop_ptr_));
+        wasm_vm_->makeGlobal("env", "DYNAMICTOP_PTR", Word(emscripten_dynamic_base_));
```
PiotrSikora: I don't think this is correct, [`DYNAMICTOP_PTR`'s address is encoded in `emscripten_dynamictop_ptr_`](https://github.com/emscripten-core/emscripten/blob/f715b7de7426bb27982c4fdeb94b5a32966781e1/tools/shared.py#L3076), and not in `emscripten_dynamic_base_`, which represents `DYNAMIC_BASE`.

Is this still necessary after your other changes? Or does this simply align with what WAVM is doing internally (which may or may not be correct, since that code predates Emscripten metadata, AFAIK).

source/extensions/common/wasm/v8/v8.cc
```diff
@@ -441,6 +443,7 @@ void V8::setMemoryLayout(uint64_t stack_base, uint64_t heap_base, uint64_t heap_
   memory_stack_base_ = stack_base;
   memory_heap_base_ = heap_base;
   memory_heap_base_pointer_ = heap_base_pointer;
+  global_dynamictop_ = makeGlobal("env", "DYNAMICTOP_PTR", Word(heap_base_pointer));
```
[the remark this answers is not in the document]
PiotrSikora: Outdated comment, nevermind.

test/extensions/wasm/config_test.cc
```diff
@@ -80,6 +80,10 @@ TEST_P(WasmFactoryTest, CreateWasmFromWASMPerThread) {
 }
 
 TEST_P(WasmFactoryTest, MissingImport) {
+  if (GetParam() == "wavm") {
+    // WAVM Emscripten support always defines stubs for all unresolved functions.
```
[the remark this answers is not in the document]
jplevyak: Yes.  We might have to change this when we move to WASI.

source/extensions/common/wasm/wasm.cc

PiotrSikora:
This fixes #213, correct? All test files are generated with 1.38.46?

source/extensions/common/wasm/wavm/BUILD
```diff
@@ -20,6 +20,7 @@ envoy_cc_library(
         '-DWASM_API=""',
         '-DWASTPARSE_API=""',
         '-DIR_API=""',
+        '-DWAVM_API=""',
```
AndrewScheidecker: FYI the WAVM_API definition supercedes all the other *_API definitions, so you should be able to remove them.

test/extensions/wasm/config_test.cc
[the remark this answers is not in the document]
AndrewScheidecker: Is this still necessary after [this commit](https://github.com/WAVM/WAVM/commit/7131af5e573779c2d2ed306baf96a103fd83b7fa )? The WAVM Emscripten resolver no longer generates stubs for unresolved functions.

source/extensions/common/wasm/wavm/BUILD
jplevyak: Updated. Also removed the missing imports check below.  Thanx!

Issue #219, opened 2019-09-28
Investigate/Mitigate memory usage for stats.


jplevyak:
There is concern that memory usage will be high and that the memory will not be reclaimed.  It seems that StatnameSet does dedup and get GC'd but StatNamePool does not dedup and does not get reclaimed.

jmarantz:
I think a more serious issue is mutex contention, whenever a string is concerted into a stat.

As long as common wasm pattern is to capture counter/gauge/histogram handles on startup it should be OK, but if anything is looked up by name during request processing there is a going to be a thead contention issue.

PR #229, opened 2019-10-04
Fix use-after-free for Stats::Scope and share the StatNameSet to save…

jplevyak:
… memory.


jplevyak:
@mandarjog

source/extensions/common/wasm/wasm.h
```diff
@@ -694,11 +695,7 @@ class Wasm : public Envoy::Server::Wasm,
   std::unique_ptr<Global<double>> global_Infinity_;
 
   // Stats/Metrics
-  // TODO(jplevyak): replace the use of Stats::StatNameSet with something more efficient.
-  // By having a separate StatNameSet per Wasm we are duplicating all the strings but
-  // avoiding locks. Consider lock-free hash tables or pre-registering stats.
-  Stats::StatNameSet stat_name_set_;
-  absl::flat_hash_map<std::string, Stats::StatName> stat_names_;
```
lambdai: is `stat_names_` never used?

source/extensions/common/wasm/wasm.h
```diff
@@ -504,7 +504,7 @@ class Wasm : public Envoy::Server::Wasm,
 public:
   Wasm(absl::string_view vm, absl::string_view vm_id, absl::string_view vm_configuration,
        PluginSharedPtr plugin, Upstream::ClusterManager& cluster_manager,
-       Event::Dispatcher& dispatcher);
+       Event::Dispatcher& dispatcher, std::shared_ptr<Stats::StatNameSet> stat_name_set = nullptr);
```
PiotrSikora: Nitpicking now, since this is going to be a problem when upstreaming anyway:
1) `std::shared_ptr<Stats::StatNameSet>` -> `Stats::StatNameSetSharedPtr`.
2) I don't think there is a convention of setting default argument values in prototypes in Envoy.

source/extensions/common/wasm/wasm.h
jplevyak: It isn't used now.

The use-after-free was caused by the constructor for the StatNameSet being called after the ListenerImpl was deleted.  That can't happen now because the StatNameSet constructor is called on the main thread and then the shared_ptr is passed to all the other threads vial the TLS Slot Set call.  The destructor doesn't access the Scope from the ListenerImpl so it is safe.

source/extensions/common/wasm/wasm.h
jplevyak: Done (SharedPtr).

Fixed the default argument.  PTAL

lambdai:
Discussed offline:
`stat_name_set_(creating_plugin_->scope_.symbolTable()` is the cause of the crash.
The fact is that the underlying symbolTable outlives `plugin_` and `scope_` so this PR bypass the ephemeral objects to create stat_name_set_ and fix the crash. 

/lgtm

jmarantz:
I've been following the StatNameSet usage in wasm from afar and I am speculating that this PR introduces a thread contention issue, as multiple distinct threads may now be accessing the same StatNameSet, calling getDynamic() which takes a lock.

Is that right?

jplevyak:
https://github.com/envoyproxy/envoy-wasm/issues/219

jplevyak:
Yes. That is the memory/cpu/contention trade-off for this fix.  Let's discuss next steps on the associated issue: #219

PiotrSikora:
@jplevyak could you rebase this on top of `master`?

PR #233, opened 2019-10-07
[WIP] Experiement Golang SDK

bianpengyuan:
This is some exploratory work I recently did. Lots of works remain, such as comment and test. Not sure if this worths putting more time into considering the WASI shifting. This is based on tinygo, which not yet fully supports reflect thus cannot do protobuf, but I think they are on the way there.

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -72,6 +72,15 @@ inline Word wasmResultToWord(WasmResult r) { return Word(static_cast<uint64_t>(r
 
 inline uint32_t convertWordToUint32(Word w) { return static_cast<uint32_t>(w.u64_); }
 
+inline Word getStringPtr(Context* context, Word string_ptr_ptr) {
+  auto ptr_str_view = context->wasmVm()->getMemory(string_ptr_ptr.u64_, 4).value();
+  int64_t ptr = 0;
+  for (int i = 0; i < 4; i++) {
+    ptr = ptr | (uint8_t(ptr_str_view[i]) << 8*i);
+  }
+  return Word(ptr);
+}
+
 // Convert a function of the form Word(Word...) to one of the form uint32_t(uint32_t...).
 template <typename F, F* fn> struct ConvertFunctionWordToUint32 {
   static void convertFunctionWordToUint32() {}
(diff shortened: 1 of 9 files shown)


source/extensions/common/wasm/wasm.cc
```diff
@@ -2211,6 +2260,16 @@ void Wasm::registerCallbacks() {
 
   _REGISTER_PROXY(setEffectiveContext);
 #undef _REGISTER_PROXY
+
+#define _REGISTER_GOLANG_PROXY(_fn)                                                                  \
```
PiotrSikora: I'm confused by why you need those? C++, Rust and TinyGo should be all calling the same functions.

bianpengyuan: tinygo imports function with package name as prefix (`envoywasmsdk`) and two extra params appended at the end: https://tinygo.org/compiler-internals/calling-convention/.

bianpengyuan: This is wat of tinygo generated wasm file:

``` 
(import "env" "envoywasmsdk._proxy_getHeaderMapValue" (func $envoywasmsdk._proxy_getHeaderMapValue (type $t2)))
  (import "env" "envoywasmsdk._proxy_log" (func $envoywasmsdk._proxy_log (type $t3)))
  (import "env" "envoywasmsdk._proxy_addHeaderMapValue" (func $envoywasmsdk._proxy_addHeaderMapValue (type $t2)))
```

ly-2016: https://github.com/tinygo-org/tinygo/pull/455

Shikugawa:
How's it going this PR?

zhongfox:
How's it going this PR?

dinusha92:
Is there any update on this? :)

PiotrSikora:
@Shikugawa @zhongfox @dinusha92 I don't believe anybody from our team is actively working on the Proxy-Wasm Go SDK (either TinyGo or regular Go), but we could provide some guidance and/or code reviews if you're interested in working on it. 

cc @bianpengyuan @incfly @jplevyak

source/extensions/common/wasm/wasm.cc
aykevl: I just came across this PR. Please note that the calling convention should be considered an implementation detail. If you want a stable calling convention, use `//export` which is intended to match the C calling convention.

mathetake:
hi folks, I personally developed go sdk for proxy-wasm : https://github.com/mathetake/proxy-wasm-go

This is still bare minimum but it works fine within the realm of basic examples without modifying envoryproxy's host implementations.

hobbytp:
Any update? thanks,

PiotrSikora:
> Any update? thanks,

You should use https://github.com/mathetake/proxy-wasm-go.

PR #239, opened 2019-10-10
build: update wee8 to 7.8.279.17.

PiotrSikora:


PiotrSikora:
@mandarjog this should fix your build issues (the fix is actually included in the archive, where `use_system_xcode` is hardcoded to `true` instead of based on the result of the script).

PR #241, opened 2019-10-10
Move WasmState into it's own interoperation lib, so that it can be used by other filters too

gargnupur:
Move WasmState into WASM Api, so that it can be used by other filters too

Description:Move WasmState into WASM Api, so that it can be used by other filters…
Risk Level: None
Testing: Existing tests will cover that
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/api/wasm/cpp/proxy_wasm_state.h
+++ b/api/wasm/cpp/proxy_wasm_state.h
@@ -8,6 +8,10 @@
 #include <string>
 
 #include "envoy/stream_info/filter_state.h"
+#include "common/protobuf/protobuf.h"
+
+namespace Envoy {
+namespace Wasm {
 
 // A simple wrapper around generic values
 class WasmState : public StreamInfo::FilterState::Object {
@@ -18,3 +22,5 @@ public:
 private:
   const ProtobufWkt::Value value_;
 };
+}
+}
--- a/api/wasm/cpp/proxy_wasm_state.h
+++ b/api/wasm/cpp/proxy_wasm_state.h
@@ -1,0 +1,20 @@
+/*
+ * Wasm State Class available to WASM modules.
+ */
+// NOLINT(namespace-envoy)
+
+#pragma once
+
+#include <string>
+
+#include "envoy/stream_info/filter_state.h"
+
+// A simple wrapper around generic values
+class WasmState : public StreamInfo::FilterState::Object {
+public:
+  WasmState(ProtobufWkt::Value& value) : value_(value) {}
+  const ProtobufWkt::Value& value() const { return value_; }
+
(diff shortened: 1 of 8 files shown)


api/wasm/cpp/proxy_wasm_state.h
```diff
@@ -0,0 +1,20 @@
...
+// NOLINT(namespace-envoy)
+
+#pragma once
+
+#include <string>
+
+#include "envoy/stream_info/filter_state.h"
+
+// A simple wrapper around generic values
+class WasmState : public StreamInfo::FilterState::Object {
```
kyessenov: I think we need some namespace for this?

gargnupur: other classes in this directory are not under a namespace.. Ex: WasmData.. should we put this under namespace?

kyessenov: WasmState should be usable in the rest of native envoy to be able to exchange data with native extensions. I think picking some Envoy-consistent namespace is necessary for it. Other stuff here I don't have a good reason to expose.

gargnupur: Added Envoy::Wasm namespace... (am open to namespace name suggestions too : )

gargnupur:
@kyessenov , @PiotrSikora, @jplevyak  : Looks like tests that failed are already failing and all the cireclci tests passed. Can you please take a look at the review?

api/wasm/cpp/proxy_wasm_state.h
jplevyak: The API should not have any namespaces.  We are trying to standardize this and it should not be Envoy specific.

jplevyak: We add the namespaces when we include the file into Envoy (see wasm.cc and the way that proxy_wasm_result.h is included).

kyessenov:
@jplevyak I wonder if it's worth having this filter state without having any common static type knowledge between the guest and the host. It's in fact just a wrapper around protobuf value. We could instead go back to dynamic metadata but not use `setDynamicMetadata` and just write the value directly into the metadata. It's effectively same as what we do with the filter state. WDYT?

jplevyak:
RE: go back to dynamic metadata  I don't know the who history, but I think the idea is to get rid of it for other reasons (i.e. for other use cases which do have better defined data structures FilterState is going to be more efficient).

So I am inclined to keep this the way it is.

gargnupur:
@jplevyak : Can you please help merge this? I don't have write access to merge...

PR #246, opened 2019-10-11
v8: fix printing values for parameters other than wasm::I32.

PiotrSikora:


source/extensions/common/wasm/v8/v8.cc
```diff
@@ -494,7 +509,8 @@ void V8::callModuleFunction(Context* context, absl::string_view function_name,
 
 void V8::callModuleFunction(Context* context, absl::string_view function_name,
                             const wasm::Func* func, const wasm::Val args[], wasm::Val results[]) {
-  ENVOY_LOG(trace, "[wasm] callModuleFunction(\"{}\")", function_name);
+  // TODO(PiotrSikora): print params when/if needed.
+  ENVOY_LOG(trace, "[wasm] [host->vm] {}({})", function_name, args ? "???" : "");
```
PiotrSikora: Note: this helper function is always* called for Wasm functions without any parameters and without any return values, so I didn't want to add complexity for pretty printing values than are never passed anyway.

*with the exception of modules compiled using Emscripten that only produced metadata version 0.0 (i.e. without memory layout), but then we don't care about those parameters, since they are hardcoded anyway.

PR #248, opened 2019-10-11
add the default maglev hash policy to be based on host/:authority

kyessenov:
Hash policy is required for consistent hash LB.

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -1593,10 +1593,16 @@ uint32_t Context::httpCall(absl::string_view cluster, const Pairs& request_heade
     token = next_http_call_token_++;
   }
   auto& handler = http_request_[token];
+
+  // set default hash policy to be based on :authority to enable consistent hash
+  Http::AsyncClient::RequestOptions options;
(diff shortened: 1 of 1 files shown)


source/extensions/common/wasm/wasm.cc
```diff
@@ -1593,10 +1593,16 @@ uint32_t Context::httpCall(absl::string_view cluster, const Pairs& request_heade
     token = next_http_call_token_++;
   }
   auto& handler = http_request_[token];
+
+  // set default hash policy to be based on :authority to enable consistent hash
+  Http::AsyncClient::RequestOptions options;
+  options.setTimeout(timeout);
+  Protobuf::RepeatedPtrField<envoy::api::v2::route::RouteAction::HashPolicy> hash_policy;
+  hash_policy.Add()->mutable_header()->set_header_name(Http::Headers::get().Host.get());
```
jplevyak: If this is the accessor you are going to use then please change the comment to be :authority/Host.

kyessenov: Updated the PR title and the comment above. Envoy conflates host and :authority, they are actually the same in the code.

PR #250, opened 2019-10-13
Replace stress test ad hoc metrics with proper Envoy stats

duderino:
No new functionality is introduced in this PR, but I'm sending it out now as a sacrifice to the gods of small pull requests.

Next step is performance metrics.

### DIFF
--- a/test/stress/stress_test_upstream.h
+++ b/test/stress/stress_test_upstream.h
@@ -198,14 +198,23 @@ public:
   void operator=(const LocalListenSocket&) = delete;
 };
 
+#define ALL_SERVER_HELPER_STATS(COUNTER)                                                           \
+  COUNTER(accepts)                                                                                 \
+  COUNTER(requests_received)                                                                       \
+  COUNTER(local_closes)                                                                            \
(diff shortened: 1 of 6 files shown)


test/stress/stress_test_upstream.h
```diff
@@ -198,50 +198,65 @@ class LocalListenSocket : public Network::TcpListenSocket {
...
   ServerCloseCallback close_callback_;
-
-  std::atomic<uint32_t> accepts_{0};
-  std::atomic<uint32_t> requests_received_{0};
-  uint32_t local_closes_{0};
-  uint32_t remote_closes_{0};
+  // Counters needed for control flow are not implemented as stats.
+  uint32_t connections_accepted_{0};
+  uint32_t connections_closed_{0};
+  std::unique_ptr<Stats> stats_;
```
jplevyak: I would put the mutex_ above the fields that it is protecting and add a GUARDED_BY annotation for those fields.

jplevyak: Add mutex_.AssertHeld() in the callback.

Issue #212, opened 2019-09-27
Switch to STANDALONE_WASM=1 to get rid of all the JS-ism in Wasm modules.


PiotrSikora:
Switch to STANDALONE_WASM=1 to get rid of all the JS-ism in Wasm modules.

Note that it requires using the "upstream" backend, so this is blocked on #77.

PiotrSikora:
@jplevyak assigning this to you, since you already started some work in #211, but feel free to reassign it to me, if you prefer that I do it.

PiotrSikora:
I spent whole day trying to get `upstream` to work with protobufs, this is far from trivial (nor urgent), so unassigning you for the time being.

PR #260, opened 2019-10-18
Use Emscripten with STANDALONE_WASM=1.

### LINKED ISSUE #77: Switch Emscripten from "fastcomp" to "upstream"

PiotrSikora:
This PR is a bit overloaded and includes a lot of changes,
but it would be impossible to split them up:

 1. Emscripten: update to 1.38.48.
 2. Emscripten: use "upstream" instead of "fastcomp".
 3. Emscripten: use STANDALONE_WASM=1.
 4. Emscripten: use __wasm_global_ctors.
 5. Emscripten: use WASI calls.
 6. Emscripten: drop underscore prefix.
 7. Emscripten: remove support for the JavaScript-flavored Emscripten.
 8. Runtimes: throw WasmException for function call errors.
 9. Tests: catch DivByZero (which is now an runtime error).
10. WAVM: update to 2019-10-15.
11. WAVM: stop using WAVM's Emscripten implementation, since we have our own.
12. SDK: distribute static libraries instead of the bytecode.

Fixes #77, #212.


PiotrSikora:
@AndrewScheidecker WAVM isn't running as part of CI, so you cannot see it here, but when updating it from 2019-10-01 (https://github.com/WAVM/WAVM/commit/6146b4e95d4b5d80d1fde9bb920791836a9078b) to 2019-10-15 (https://github.com/WAVM/WAVM/commit/1ec06cd202a922015c9041c5ed84f875453c4dc7), one of our tests started failing because `tryCollectCompartment(std::move(compartment_))` started returning `false` in a test where WAVM compartment is cloned.

I've tried changing existing code:
```
std::unique_ptr<WasmVm> Wavm::clone() {
  auto wavm = std::make_unique<Wavm>();
  wavm->compartment_ = WAVM::Runtime::cloneCompartment(compartment_);
  wavm->memory_ = WAVM::Runtime::remapToClonedCompartment(memory_, wavm->compartment_);
  wavm->memory_base_ = WAVM::Runtime::getMemoryBaseAddress(wavm->memory_);
  wavm->context_ = WAVM::Runtime::createContext(wavm->compartment_);
  for (auto& p : intrinsic_module_instances_) {
    wavm->intrinsic_module_instances_.emplace(
        p.first, WAVM::Runtime::remapToClonedCompartment(p.second, wavm->compartment_));
  }
  wavm->module_instance_ =
      WAVM::Runtime::remapToClonedCompartment(module_instance_, wavm->compartment_);
  return wavm;
}
```
to clone the context instead, i.e.:
```
 std::unique_ptr<WasmVm> Wavm::clone() {
   auto wavm = std::make_unique<Wavm>();
   wavm->compartment_ = WAVM::Runtime::cloneCompartment(compartment_);
+  wavm->context_ = WAVM::Runtime::cloneContext(context_, wavm->compartment_);
   wavm->memory_ = WAVM::Runtime::remapToClonedCompartment(memory_, wavm->compartment_);
   wavm->memory_base_ = WAVM::Runtime::getMemoryBaseAddress(wavm->memory_);
-  wavm->context_ = WAVM::Runtime::createContext(wavm->compartment_);
   for (auto& p : intrinsic_module_instances_) {
     wavm->intrinsic_module_instances_.emplace(
         p.first, WAVM::Runtime::remapToClonedCompartment(p.second, wavm->compartment_));
```
but that didn't make any difference.

I've looked at all the commits, but nothing obvious jumps out... any ideas?

cc @jplevyak

AndrewScheidecker:
> I've looked at all the commits, but nothing obvious jumps out... any ideas?

I think what must be happening is:
* Your initial `Wavm` object that you used as the source for the clone is still around, along with its `Emscripten::Instance`.
* I added some `GCPointer<Function>` to `Emscripten::Instance` [here](https://github.com/WAVM/WAVM/commit/5bb1b2ff5e68598c76f457cf3aba4d5f375d86f8#diff-f4c157ae28abbc434da411671e2f0d23R102).
* The WAVM `Function` objects are strange objects that may be shared between cloned compartments: they are just pointers to a header that is emitted before the machine code for the function.
* The `GCPointer<Function>` is treated as a GC root in *all* compartments that contain the function, and so none of the compartments that were cloned from the original can be collected until the `GCPointer<Function>` in the `Emscripten::Instance` is destroyed.

I should perhaps get rid of the `GCPointer<Function>` in `Emscripten::Instance`, but I think that there are likely other problems lurking here due to the `Emscripten::Instance` being shared between cloned compartments. To solve this I could add an API like this:
```
std::shared_ptr<Emscripten::Instance> remapToClonedCompartment(Emscripten::Instance&,
                                                               Compartment*)
```

However, I am wondering if you need the `Emscripten::Instance` at all. It looks like you have your own Emscripten emulation layer, so is it possible to remove WAVM's?

PiotrSikora:
> However, I am wondering if you need the `Emscripten::Instance` at all. It looks like you have your own Emscripten emulation layer, so is it possible to remove WAVM's?

That's an excellent question! We don't need it anymore, and removing it fixes the issue.

Thanks a lot!

api/wasm/cpp/sdk_container.sh
```diff
@@ -1,13 +1,15 @@
...
+
 # basics
 export DEBIAN_FRONTEND=noninteractive
 apt-get update
 apt-get upgrade -y
 apt-get install -y --no-install-recommends apt-utils ca-certificates
 apt-get autoremove -y
 apt-get clean
-apt-get install -y --no-install-recommends software-properties-common apt-transport-https git wget curl libglib2.0-dev autoconf autotools-dev automake libtool cmake python
+apt-get install -y --no-install-recommends software-properties-common apt-transport-https git wget curl pkg-config autoconf autotools-dev automake libtool cmake python zlib1g-dev
```
jplevyak: Just curious what caused these changes?

source/extensions/common/wasm/null/null_vm.h
```diff
@@ -29,8 +29,7 @@ struct NullVm : public WasmVm {
   bool cloneable() override { return true; };
   WasmVmPtr clone() override;
   bool load(const std::string& code, bool allow_precompiled) override;
-  void link(absl::string_view debug_name, bool needs_emscripten) override;
-  void setMemoryLayout(uint64_t, uint64_t, uint64_t) override {}
+  void link(absl::string_view debug_name) override;
```
jplevyak: We are going to have to upstream this.

PiotrSikora: Yup, V8 isn't merged yet, so it's fine I think.

api/wasm/cpp/sdk_container.sh
PiotrSikora: I removed `libglib2.0-dev` which wasn't necessary (it used to be needed by V8, but it's not anymore, and I'm not sure why it was here in the first place), so I had to add `pkg-config` and `zlib1g-dev` directly, since those are needed and were previously installed as transitive dependencies of `libglib2.0-dev`.

PR #261, opened 2019-10-18
Add access to streaminfo for tcp connections too

gargnupur:
Description: Add access to streaminfo for tcp connections too
Risk Level: Low

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -1590,8 +1590,6 @@ StreamInfo::StreamInfo* Context::getRequestStreamInfo() const {
     return &encoder_callbacks_->streamInfo();
   } else if (decoder_callbacks_) {
     return &decoder_callbacks_->streamInfo();
-  } else if (access_log_stream_info_) {
-    return access_log_stream_info_;
   } else if (network_read_filter_callbacks_) {
     return &network_read_filter_callbacks_->connection().streamInfo();
   } else if (network_write_filter_callbacks_) {
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -1577,6 +1577,10 @@ const StreamInfo::StreamInfo* Context::getConstRequestStreamInfo() const {
     return &decoder_callbacks_->streamInfo();
   } else if (access_log_stream_info_) {
     return access_log_stream_info_;
+  } else if (network_read_filter_callbacks_) {
+    return &network_read_filter_callbacks_->connection().streamInfo();
+  } else if (network_write_filter_callbacks_) {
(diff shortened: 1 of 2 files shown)


source/extensions/common/wasm/wasm.cc
```diff
@@ -1722,21 +1722,38 @@ const StreamInfo::StreamInfo* Context::getRequestStreamInfo() const {
...
+  if (encoder_callbacks_) {
+    return encoder_callbacks_->streamInfo().filterState();
+  }
+  if (decoder_callbacks_) {
+    return decoder_callbacks_->streamInfo().filterState();
+  }
+  if (network_read_filter_callbacks_) {
+    return network_read_filter_callbacks_->connection().streamInfo().filterState();
+  }
+  return network_write_filter_callbacks_->connection().streamInfo().filterState();
```
PiotrSikora: I don't think this is correct, you have no guarantee that this isn't `nullptr`.

Also, does `network_read_filter_callbacks_->connection()` and `network_write_filter_callbacks_->connection()` point to the same connection or is one for downstream and one for upstream?

source/extensions/common/wasm/wasm.cc
```diff
@@ -1722,21 +1722,38 @@ const StreamInfo::StreamInfo* Context::getRequestStreamInfo() const {
...
+}
+
 WasmResult Context::setProperty(absl::string_view key, absl::string_view serialized_value) {
   ProtobufWkt::Value value_struct;
   if (!value_struct.ParseFromArray(serialized_value.data(), serialized_value.size())) {
     return WasmResult::ParseFailure;
   }
-  if (decoder_callbacks_ == nullptr && encoder_callbacks_ == nullptr) {
```
PiotrSikora: Everything below could be simplified to:
```
auto stream_info = getRequestStreamInfo();
if (!stream_info) {
  return WasmResult:: NotFound;
}

stream_info->filterState().setData(key, std::make_unique<WasmState>(value_struct),
                                   StreamInfo::FilterState::StateType::Mutable);
return WasmResult::Ok;
```
and then you don't even need `getFilterState()`.

gargnupur: I totally agree with that, but I kept separate functions because getStreamInfo was looking at access_log_stream_info_  and getFilterstate was not..  Is it okay to look at access_log_stream_info_ for filter state?

source/extensions/common/wasm/wasm.cc
gargnupur: Thanks for pointing..will fix that..

looks like they both point to the same connection.. we need different one for upstream and downstream right?

gargnupur:
@PiotrSikora : Can you please take a look at this?

PR #264, opened 2019-10-19
Reduce debug information in generated .wasm files.

PiotrSikora:
This reduces the size of .wasm files by 20 to 40%.


jplevyak:
It would be great if we could get symbol information in backtraces out of WASM.  Should we open an issue to work with the V8 and emscripten folks to see if we can do that?

PiotrSikora:
Well, this PR is kind of doing the opposite... 1.39.0 accidentally started adding `debug_info` section with source maps when using `-g3` (which should be only added at `-g4`), which increased the `.wasm` size 3-4x, so I took the opportunity and completely removed `-g` flag, since we don't need it anymore (Emscripten used to mangle functions names below `-g3`, but that's fixed now).

Regarding stacktraces, I was already started playing with that, but it seems that we need some way to combine offsets with source maps, see: #266.

PR #267, opened 2019-10-21
Remove support for __errno_location.

PiotrSikora:
We don't need it anymore. Missed in #260.


PiotrSikora:
There are no lib support functions anymore. With `STANDALONE_WASM`, only a few WASI functions are exported from the host and they use `__wasi_errno_t` return values. Everything else is handled within the WASM module itself (thanks to musl being compiled in, without relying on JS bits).

`__errno_location` is not even exported anymore, so this is effectively a dead code anyway...

PR #268, opened 2019-10-21
Move start-up logic to the abstraction layer.

PiotrSikora:


source/extensions/common/wasm/wasm.cc
```diff
@@ -2240,12 +2243,21 @@ bool Wasm::initialize(const std::string& code, bool allow_precompiled) {
...
-  wasm_vm_->start(vm_context_.get());
+  startVm(vm_context_.get());
   code_ = code;
   allow_precompiled_ = allow_precompiled;
   return true;
 }
 
+void Wasm::startVm(Context* root_context) {
+  /* Call "_start" function, and fallback to "__wasm_call_ctors" if the former is not available. */
+  if (_start_) {
```
jplevyak: We used to call both of these if there were available.  So _start calls __wasm_call_ctors ?

jplevyak: See https://github.com/rustwasm/wasm-bindgen/issues/1216

PiotrSikora: Yes, `_start` calls `__wasm_call_ctors`, so we don't want call it twice.

But we never called both (`_start` was added in `1.39.0`). We used to look for a "start" function, but none of the toolchains (Emscripten, WASI, Rust) use this WebAssembly indicator, and it's not supported by V8 anyway.

PiotrSikora: We don't use wasm-bindgen (it's for JavaScript<>Rust bindings).

PR #271, opened 2019-10-22
Change the ownership and lifetime of the Scope used by Wasm for Stats.

jplevyak:
Change the ownership of the Scope to shared and change createScope with "" to not add a level of ".".  This is to address the segv resulting from the Scope stored in the Listener being destroyed.


kyessenov:
Let's get this in. Do we lose any stats at all? Is it just some gRPC call counters?

PR #272, opened 2019-10-23
Cancel outstanding operations when the controlling Context is destroyed.

jplevyak:

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -2045,6 +2045,23 @@ std::string Plugin::makeLogPrefix() const {
   return prefix;
 }
 
+Context::~Context() {
+  // Cancel any outstanding requests.
+  for (auto& p : http_request_) {
+    p.second.request->cancel();
+  }
+  for (auto& p : grpc_call_request_) {
+    p.second.request->cancel();
+  }
+  for (auto& p : grpc_stream_) {
+    p.second.stream->resetStream();
+  }
+  // Do not remove vm or root contexts which have the same lifetime as wasm_.
+  if (root_context_id_) {
+    wasm_->contexts_.erase(id_);
+  }
+}
+
 void Wasm::registerCallbacks() {
 #define _REGISTER(_fn)                                                                             \
   wasm_vm_->registerCallback(                                                                      \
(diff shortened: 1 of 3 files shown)


source/extensions/common/wasm/wasm.cc
```diff
@@ -2045,6 +2045,23 @@ std::string Plugin::makeLogPrefix() const {
...
+Context::~Context() {
+  // Cancel any outstanding requests.
+  for (auto& p : http_request_) {
+    p.second.request->cancel();
+  }
+  for (auto& p : grpc_call_request_) {
+    p.second.request->cancel();
+  }
+  for (auto& p : grpc_stream_) {
+    p.second.stream->resetStream();
```
kyessenov: Are these all synchronous? If it posts back to libevent, the this won't help.

kyessenov: @PiotrSikora Isn't google gRPC still running in a separate silo and cancellation is async? I don't remember what happened to it afterwards, but it used to run on a separate worker thread.

test/extensions/filters/http/wasm/wasm_filter_test.cc
```diff
@@ -255,6 +255,51 @@ TEST_P(WasmHttpFilterTest, AsyncCall) {
...
+  Http::MessagePtr response_message(new Http::ResponseMessageImpl(
+      Http::HeaderMapPtr{new Http::TestHeaderMapImpl{{":status", "200"}}}));
+  response_message->body().reset(new Buffer::OwnedImpl("response"));
+
+  // (Don't) Make the callback on the destroyed VM.
+  EXPECT_EQ(callbacks, nullptr);
+  if (callbacks) {
+    callbacks->onSuccess(std::move(response_message));
+  }
+}
```
PiotrSikora: Actually, could you add tests for all the types (HTTP callouts, gRPC callouts and gRPC streams)?

source/extensions/common/wasm/wasm.cc
kyessenov: I checked gRPC code. LG.

test/extensions/filters/http/wasm/wasm_filter_test.cc
jplevyak: Added a gRPC one.  We currently don't have stream tests.  Need to fix that: https://github.com/envoyproxy/envoy-wasm/issues/273

That said, these tests are almost completely useless.  This issue is a threading/config interaction issue and these tests mock out so much of the system as to essentially test nothing but that the lines I added do what they say they do.

We need integration tests for all these and I would contend we should remove the mock tests since they give a false sense of security without actually testing anything that would not be better tested by the integration tests.

PR #274, opened 2019-10-25
Fix ABI: all data is pulled by the code in the VM.

jplevyak:


api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -180,8 +180,8 @@ class GrpcCallHandlerBase {
   void cancel();
 
   virtual void onCreateInitialMetadata() {}
-  virtual void onSuccess(std::unique_ptr<WasmData> message) = 0;
-  virtual void onFailure(GrpcStatus status, std::unique_ptr<WasmData> error_message) = 0;
+  virtual void onSuccess(size_t body_size) = 0;
+  virtual void onFailure(GrpcStatus status) = 0;
```
PiotrSikora: Shouldn't we include `status_code` here as well?

api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -293,60 +280,58 @@ class RootContext : public ContextBase {
...
   // Called when the timer goes off.
   virtual void onTick() {}
   // Called when data arrives on a SharedQueue.
   virtual void onQueueReady(uint32_t /* token */) {}
 
   // Low level HTTP/gRPC interface.
-  virtual void onHttpCallResponse(uint32_t token, std::unique_ptr<WasmData> header_pairs,
-                                  std::unique_ptr<WasmData> body,
-                                  std::unique_ptr<WasmData> trailer_pairs);
+  virtual void onHttpCallResponse(uint32_t token, size_t body_size);
```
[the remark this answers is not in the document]
jplevyak: Because the call to get the body requires the length requested, so it would require an extra call to get the length of the buffer before getting the whole body.

source/extensions/common/wasm/wasm.cc
```diff
@@ -679,26 +664,62 @@ Word getHeaderMapSizeHandler(void* raw_context, Word type, Word result_ptr) {
...
+  }
+  if (!context->wasmVm()->setWord(size_ptr.u64_, Word(length.u64_))) {
+    return wasmResultToWord(WasmResult::InvalidMemoryAccess);
+  }
   return wasmResultToWord(WasmResult::Ok);
 }
 
-Word getResponseBodyBufferBytesHandler(void* raw_context, Word start, Word length, Word ptr_ptr,
-                                       Word size_ptr) {
+Word getBufferStatusHandler(void* raw_context, Word type, Word length_ptr, Word flags_ptr) {
```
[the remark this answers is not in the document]
jplevyak: I think it is a necessary API for more complex codes, e.g. transcoding where otherwise all the code would be required to track and pass the size of the buffer.  With this call it is possible to transfer control of the stream without being required to pass that additional piece of information in all cases.

api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -427,6 +412,21 @@ struct RegisterContextFactory {
...
 };
 
+inline WasmDataPtr getConfiguration() {
+  const char* value_ptr = nullptr;
+  size_t value_size = 0;
+  CHECK_RESULT(proxy_getConfiguration(&value_ptr, &value_size));
+  return std::make_unique<WasmData>(value_ptr, value_size);
+}
+
+inline std::pair<uint32_t, WasmDataPtr> getStatus() {
```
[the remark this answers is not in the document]
jplevyak: And for generic WasmResult != Ok.  It would be nice to have an addition details message available for debugging.

I see this more googl3 Status which is completely generic. It has error namespaces, perhaps you would be interested in adding that?  e.g. GrpcErrors, HttpErrors, WasmErrors ?

source/extensions/common/wasm/wasm.cc
```diff
@@ -2534,16 +2617,19 @@ void Context::setEncoderFilterCallbacks(Envoy::Http::StreamEncoderFilterCallback
...
 }
 
-void Context::onHttpCallFailure(uint32_t token, Http::AsyncClient::FailureReason /* reason */) {
-  onHttpCallResponse(token, {}, "", {});
+void Context::onHttpCallFailure(uint32_t token, Http::AsyncClient::FailureReason reason) {
+  status_code_ = static_cast<uint32_t>(WasmResult::BrokenConnection);
+  // This is the only value currently.
+  ASSERT(reason == Http::AsyncClient::FailureReason::Reset);
+  status_message_ = "reset";
+  onHttpCallResponse(token, 0);
```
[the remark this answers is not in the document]
jplevyak: Would that be the Http code ?

api/wasm/cpp/proxy_wasm_api.h
jplevyak: You mean on success?  It would always be Ok

jplevyak:
Updated WRT comments.  PTAL

api/envoy/config/wasm/v3alpha/wasm.proto
```diff
@@ -39,6 +37,8 @@ message VmConfig {
 message PluginConfig {
   // A unique name for a filters/services in a VM for use in identifiying the filter/service if
   // multiple filters/services are handled by the same vm_id and root_id and for logging/debugging.
+  // If left blank the plugin will be assigned an integer name corresponding to when it registers
+  // with the VM (i.e. the first plugin to register with the VM will be "0").
```
PiotrSikora: Why? Is it only to identify them? UUID might be better, then, and it's already used for that purpose in Envoy.

api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -248,12 +243,7 @@ class GrpcStreamHandler : public GrpcStreamHandlerBase {
     local_close_ = local_close_ || end_of_stream;
   }
 
-  virtual void onReceive(Response&& message) = 0;
-
-private:
-  void onReceive(std::unique_ptr<WasmData> message) override {
-    onReceive(message->proto<Response>());
-  }
+  virtual void onReceive(size_t body_size) = 0;
```
PiotrSikora: Nit: `message_size`?

api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -271,10 +261,8 @@ class ContextBase {
   virtual Context* asContext() { return nullptr; }
 
   using HttpCallCallback =
-      std::function<void(std::unique_ptr<WasmData> header_pairs, std::unique_ptr<WasmData> body,
-                         std::unique_ptr<WasmData> trailer_pairs)>;
-  using GrpcSimpleCallCallback =
-      std::function<void(GrpcStatus status, std::unique_ptr<WasmData> message)>;
+      std::function<void(uint32_t, size_t, uint32_t)>; // headers, body_size, trailers
```
PiotrSikora: `headers` / `trailers` - is that a number of items or total length when serialized?

source/extensions/common/wasm/wasm.cc
```diff
@@ -55,7 +57,6 @@ namespace Wasm {
 
 namespace {
 
-// TODO: move to utils during upstreaming.
```
PiotrSikora: Why did you remove this?

api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -179,9 +179,9 @@ class GrpcCallHandlerBase {
 
   void cancel();
 
-  virtual void onCreateInitialMetadata() {}
-  virtual void onSuccess(std::unique_ptr<WasmData> message) = 0;
-  virtual void onFailure(GrpcStatus status, std::unique_ptr<WasmData> error_message) = 0;
+  virtual void onCreateInitialMetadata(uint32_t /* headers */) {}
+  virtual void onSuccess(size_t body_size) = 0;
+  virtual void onFailure(GrpcStatus status) = 0;
```
PiotrSikora: Should this include `status_message_size`?

api/wasm/cpp/proxy_wasm_api.h
PiotrSikora: Actually, this is fine.

api/wasm/cpp/proxy_wasm_api.h
jplevyak: number

api/wasm/cpp/proxy_wasm_api.h
jplevyak: We could.  My feeling however is that this is an exceptional situation and if you have an error and need to the details, then an extra call will not kill you.

api/envoy/config/wasm/v3alpha/wasm.proto
jplevyak: Removed it.  This is purely user-facing so if they don't add one then they done need it.

source/extensions/common/wasm/wasm.cc
jplevyak: typo, added back.

PR #278, opened 2019-10-28
Clean builds in the the docker image after installation.

jplevyak:


jplevyak:
For WAVM and protobuf that is probably the case.  For abseil it may be better to compile and link individual files as the toolchain seems to be linking in the full archive currently.

jplevyak:
Done

PR #284, opened 2019-10-31
getProperty: cleanup

kyessenov:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -38,6 +38,7 @@
 #include "eval/eval/field_backed_list_impl.h"
 #include "eval/eval/field_backed_map_impl.h"
 #include "eval/public/cel_value.h"
+#include "eval/public/value_export_util.h"
 #include "openssl/bytestring.h"
 #include "openssl/hmac.h"
 #include "openssl/sha.h"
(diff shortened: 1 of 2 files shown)


source/extensions/common/wasm/wasm.cc
```diff
@@ -1059,9 +985,15 @@ WasmResult serializeValue(Filters::Common::Expr::CelValue value, std::string* re
     result->assign(reinterpret_cast<const char*>(&out), sizeof(absl::Time));
     return WasmResult::Ok;
   }
+  case CelValue::Type::kMessage:
+    if (value.MessageOrDie() != nullptr && value.MessageOrDie()->SerializeToString(result)) {
+      return WasmResult::Ok;
+    }
+    return WasmResult::SerializationFailure;
```
PiotrSikora:
```
auto msg = value.MessageOrDie();
if (msg == nullptr) {
    return WasmResult::SerializationFailure;
}
msg->SerializeToString(result);
return WasmResult::Ok;
```

But if `MessageOrDie()` is crashing the process (the name suggest that it does), then this could be even further simplified as simply:
```
value.MessageOrDie()->SerializeToString(result);
return WasmResult::Ok;
```

kyessenov: `nullptr` is a valid value here so I changed it to treat it as Ok value with empty result.

PR #293, opened 2019-11-07
merge envoy master

lizan:


PiotrSikora:
We're switching to stable releases, so this is going to remain at 1.12.x for a while...

lizan:
@PiotrSikora that's for the sha that istio-proxy pointing to right? it shouldn't block master?

lizan:
will wait on #292

jplevyak:
At this stage we do not want to be backporting and instead focus on upstreaming.

rshriram:
How long is the upstreaming going to take? If its a few days, then okay. If it is anything more than that, then that doesn't work. Because the fix that lizan is trying to merge is a bug fix for Istio (about ALS metadata logging) as it is currently logging incorrect stuff.

lizan:
@PiotrSikora @jplevyak I don't think we agreed locking down master to Envoy stable release, we agreed on locking down Istio release to Envoy stable release, to reduce maintenance cost.

lizan:
If we think #297 is ok to cherry-pick the change for bug fix, I'm fine with that too. @rshriram @PiotrSikora

PR #294, opened 2019-11-07
Add logging and stats for Wasm and WasmVm objects.

jplevyak:


source/extensions/common/wasm/null/null.cc
```diff
@@ -18,7 +18,7 @@ namespace Common {
 namespace Wasm {
 namespace Null {
 
-WasmVmPtr createVm() { return std::make_unique<NullVm>(); }
+WasmVmPtr createVm(Stats::ScopeSharedPtr scope) { return std::make_unique<NullVm>(scope); }
```
lizan: nit: `std::move(scope)` or make param const ref (same for places below)

source/extensions/common/wasm/wasm.cc
```diff
@@ -2057,13 +2058,21 @@ WasmResult Context::getMetric(uint32_t metric_id, uint64_t* result_uint64_ptr) {
...
            Stats::ScopeSharedPtr scope, Upstream::ClusterManager& cluster_manager,
            Event::Dispatcher& dispatcher)
-    : vm_id_(std::string(vm_id)), wasm_vm_(Common::Wasm::createWasmVm(vm)), scope_(scope),
-      cluster_manager_(cluster_manager), dispatcher_(dispatcher),
+    : vm_id_(std::string(vm_id)), wasm_vm_(Common::Wasm::createWasmVm(runtime, scope)),
+      scope_(scope), cluster_manager_(cluster_manager), dispatcher_(dispatcher),
       time_source_(dispatcher.timeSource()), vm_configuration_(vm_configuration),
-      stat_name_set_(scope_->symbolTable().makeSet("Wasm").release()) {}
+      wasm_stats_(WasmStats{ALL_WASM_STATS(
+          POOL_COUNTER_PREFIX(*scope_, std::string("wasm.") + std::string(runtime) + "."),
```
lizan: `absl::StrCat` instead of `+` string

source/extensions/common/wasm/null/null.cc
jplevyak: I was wondering about how effective copy elision had gotten so I ran a simple test:

struct Foo {
  Foo(int x) : x_(x) {}
  int x_;
};

struct Bar {
  Bar(std::shared_ptr<Foo> x) : x_(std::move(x)) {}
  std::shared_ptr<Foo> x_;
};

int main() {
  auto x = std::make_shared<Foo>(1);
  Bar b(std::move(x));
  printf("%d\n", b.x_->x_);
}

************ code 1 *******************

struct Foo {
  Foo(int x) : x_(x) {}
  int x_;
};

struct Bar {
  Bar(std::shared_ptr<Foo> x) : x_(x) {}
  std::shared_ptr<Foo> x_;
};

int main() {
  auto x = std::make_shared<Foo>(1);
  Bar b(x);
  printf("%d\n", b.x_->x_);
}

******************* code 2 *********************

These two codes return the exact same file when compiled:

clang++ (version 9) -O3 -S

source/extensions/common/wasm/wasm.cc
jplevyak: Done here and in the other location with the same code.

source/extensions/common/wasm/null/null.cc
jplevyak: Half this change can be upstreamed, but the wasm.cc part can't be (yet).

I will push the upstream of the applicable part once this goes in.

lizan: either is fine, but I guess clang-tidy may flag you there: https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html

PR #295, opened 2019-11-07
Move wasm Handlers to separate files (part of cleanup for upstreaming).

jplevyak:


jplevyak:
If we remove the Handler suffix then we are likely to get name collisions.  Any function which we want to export would have to have the same name.   How about putting these in a new namespace?  namespace Exports ?  We could also change the name of the file to wasm_exports.cc

jplevyak:
Done. PTAL

Issue #298, opened 2019-11-08
The RootContext plugin_name getProperty response should match the caller for onStart and onDelete and return WasmResult::NotFound otherwise.


jplevyak:
https://docs.google.com/document/d/1NR7UvS-pGUz6FQ4IHqQ5weHjQAOWQdeKIUjKtPEoJd0/edit#

PR #299, opened 2019-11-11
In accordance with the Proxy WebAssembly Lifecycle doc,

jplevyak:
root context calls onStart and onConfigure have the plugin context
corresponding to the plugin which initiated the call while
generic non-pluing sourced calls such as onHttpCallResponse will return
WasmResult::NotFound for plugin specific data.


jplevyak:
https://github.com/envoyproxy/envoy-wasm/issues/298

PR #301, opened 2019-11-12
sync bazelrc from upstream

lizan:


PiotrSikora:
Nit: It might have been a good idea to wait for the CI to complete before merging.

PR #310, opened 2019-11-14
Use connection's filter state object when available for filter state

gargnupur:
Use connection's filter state object when available as compared to request's filter state object

Description: Use connection's filter state object when available as compared to request's filter state object
Risk Level: Medium
Testing: Test in Istio/Proxy
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -344,10 +344,11 @@ WasmResult serializeValue(Filters::Common::Expr::CelValue value, std::string* re
 class WasmStateWrapper : public google::api::expr::runtime::CelMap {
 public:
   WasmStateWrapper(const StreamInfo::FilterState& filter_state,
-                     const StreamInfo::FilterState* connection_filter_state)
-        : filter_state_(filter_state), connection_filter_state_(connection_filter_state) {}
+                   const StreamInfo::FilterState* connection_filter_state)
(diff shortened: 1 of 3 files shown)


gargnupur:
> We should not displace request filter state. These are different filter states -- one applies to many requests. I think we need to expose both for now using a different accessor, e.g. `connection_filter_state`.
> We will need upstream filter state later, too.

@kyessenov : thanks for the feedback. Yes, I saw stats plugin stopped working because of this logic. Added accessor for connection_filter_state.

gargnupur:
@kyessenov , @PiotrSikora ,@jplevyak: This is ready for review..

source/extensions/common/wasm/context.cc
```diff
@@ -427,8 +441,13 @@ WasmResult Context::getProperty(absl::string_view path, std::string* result) {
         value = CelValue::CreateMessage(&info->dynamicMetadata(), &arena);
         break;
       case PropertyToken::FILTER_STATE:
-        value = CelValue::CreateMap(
-            Protobuf::Arena::Create<WasmStateWrapper>(&arena, info->filterState()));
+        if (getConnection()) {
```
jplevyak: Store the connection in a local and check if it is nullptr then reuse that local to get the filter state.

PR #311, opened 2019-11-14
Add RootContext life cycle of onDone and onDelete.

jplevyak:


test/extensions/wasm/wasm_test.cc
```diff
@@ -77,24 +77,32 @@ TEST_P(WasmTestMatrix, Logging) {
       absl::StrCat("envoy.wasm.runtime.", std::get<0>(GetParam())), vm_id, vm_configuration, scope,
       cluster_manager, *dispatcher);
   EXPECT_NE(wasm, nullptr);
+  auto wasm_ptr = wasm.get();
+  auto wasm_handler = std::make_unique<Extensions::Common::Wasm::WasmHandle>(std::move(wasm));
```
PiotrSikora: I know this is only a test, but we're giving away the `wasm` object here and are referring to its raw pointer in rest of the code, even after `wasm_handler.reset()`, at which points we have no guarantee that it still point to anything valid.

jplevyak: In this test it has to be valid.  Let me do it with a weak_ptr.  That will assert in an appropriate way if the test fails.

jplevyak: done

PiotrSikora: I was thinking more of simply moving the `std::move` to few lines below, but that works as well. Thanks!

test/extensions/wasm/wasm_test.cc
```diff
@@ -73,28 +73,37 @@ TEST_P(WasmTestMatrix, Logging) {
...
+  wasm_weak.lock()->setContext(context.get());
   auto root_context = context.get();
-  wasm->startForTesting(std::move(context), plugin);
-  wasm->configure(root_context, plugin, "configure-test");
-  wasm->tickHandler(root_context->id());
+  wasm_weak.lock()->startForTesting(std::move(context), plugin);
+  wasm_weak.lock()->configure(root_context, plugin, "configure-test");
+  wasm_handler.reset();
+  dispatcher->run(Event::Dispatcher::RunType::NonBlock);
+  // This will SEGV on nullptr if wasm has been deleeted.
```
PiotrSikora: Nit: `s/deleeted/deleted/`

jplevyak: done

PR #313, opened 2019-11-15
Make separate context.h context.cc for Wasm Context.

jplevyak:
This follows the wasm-done PR https://github.com/envoyproxy/envoy-wasm/pull/311 and includes those changes.  See the last 2 commits for differences or wait for the merge.


PiotrSikora:
Two high-level comments:
1) I still think we should split the interfaces further into filters (i.e. what I did here: https://github.com/envoyproxy/envoy-wasm/pull/171/commits/cc79026b26c3dac52f40bdaaee4fbf52868b4d84).
2) Perhaps we should use `WasmState` instead of `Context` to match Lua filters?

jplevyak:
> 1. I still think we should split the interfaces further into filters (i.e. what I did here: [cc79026](https://github.com/envoyproxy/envoy-wasm/commit/cc79026b26c3dac52f40bdaaee4fbf52868b4d84)).

I was thinking about going a slightly different route and instead having separate HttpFilterContext and NetworkFilterContext which are all mix-ins of the base Context and provide the Envoy filter interfaces while the Context still implements all of the handlers.  This doesn't add another object which makes testing easier, compartmentalizes the Envoy APIs and still makes it easy to share Context implementations (e.g. getBufferByte).

> 2. Perhaps we should use `WasmState` instead of `Context` to match Lua filters?

That is an artifact of the fact that Lua uses co-routines, and so has state rather than context.  gRPC is a more appropriate comparative, which uses context (ctx).  If we move to co-routines then I would agree.  clang-9 has early c++ co-routine support.  At some point I'd like to see what such an API would look like.

jplevyak:
I would like to get this in as a code reorganization PR and then do renaming and other reorganization (e.g. your high level comments) in subsequent PRs.

PiotrSikora:
Sure, sounds good. Please merge the other PR and merge the changes.

jplevyak:
Done.  PTAL

PR #314, opened 2019-11-18
wasm: make V8-based runtime cloneable.

### LINKED ISSUE #161: Make wee8-based VM cloneable.

PiotrSikora:
Fixes #161.

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -1608,7 +1608,7 @@ bool Wasm::initialize(const std::string& code, bool allow_precompiled) {
     allow_precompiled_ = allow_precompiled;
   }
 
-  if (started_from_ < Cloneable::InstantiatedModule) {
+  if (started_from_ != Cloneable::InstantiatedModule) {
     registerCallbacks();
     wasm_vm_->link(vm_id_);
   }
@@ -1616,7 +1616,7 @@ bool Wasm::initialize(const std::string& code, bool allow_precompiled) {
   vm_context_ = std::make_shared<Context>(this);
   getFunctions();
 
-  if (started_from_ < Cloneable::InstantiatedModule) {
+  if (started_from_ != Cloneable::InstantiatedModule) {
     // Base VM was already started, so don't try to start cloned VMs again.
     startVm(vm_context_.get());
   }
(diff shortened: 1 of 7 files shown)


source/extensions/common/wasm/wasm.cc
```diff
@@ -1407,24 +1407,20 @@ void Wasm::registerCallbacks() {
       "env", #_fn, &Exports::_fn,                                                                  \
       &ConvertFunctionWordToUint32<decltype(Exports::_fn),                                         \
                                    Exports::_fn>::convertFunctionWordToUint32)
-  if (is_emscripten_) {
-    _REGISTER(pthread_equal);
-  }
+  _REGISTER(pthread_equal);
```
PiotrSikora: Note: I remove the `is_emscripten_` guard here and for WASI because clones don't have access to Wasm code, and we don't copy this value to them... and while we could set `is_emscripten_` in clones, seeing as WASI is becoming a standard and other toolchains use it as well, I think it's fine for us to always expose those functions, so that modules can link against them, if needed.

source/extensions/common/wasm/wasm.cc
```diff
@@ -1571,50 +1565,62 @@ bool Wasm::initialize(const std::string& code, bool allow_precompiled) {
     return false;
   }
 
-  // Construct a unique identifier for the VM based on the provided vm_id and a hash of the
-  // code.
-  vm_id_with_hash_ = vm_id_ + ":" + base64Sha256(code);
+  if (started_from_ == Cloneable::NotCloneable) {
```
PiotrSikora: This is just a bunch of extra `if (started_from_ ...)` guards and indents, but GitHub somehow didn't catch that, so it's best to review locally with `git show -w`.

source/extensions/common/wasm/wasm.cc
jplevyak: Agreed

source/extensions/common/wasm/wasm.cc
```diff
@@ -1571,50 +1565,62 @@ bool Wasm::initialize(const std::string& code, bool allow_precompiled) {
...
+          start = decodeVarint(start, end, &emscripten_standalone_wasm_);
+        }
       }
     }
+
+    code_ = code;
+    allow_precompiled_ = allow_precompiled;
+  }
+
+  if (started_from_ < Cloneable::InstantiatedModule) {
```
jplevyak: != instead of < as that provides more information in this context and is equivalent.

jplevyak: And below

PiotrSikora: Done.

jplevyak:
I am going to assume you are good with me merging this.  I want to upstream the latest V8 code before I upstream the first ABI bit.

PiotrSikora:
I have AOT almost ready, and I wanted to send both at the same time to upstream.

PR #316, opened 2019-11-20
wasm: support precompiled modules in V8-based runtime.

PiotrSikora:


source/extensions/common/wasm/v8/v8.cc
```diff
@@ -241,15 +252,33 @@ template <typename T, typename U> constexpr T convertValTypesToArgsTuple(const U
...
-  module_ = wasm::Module::make(store_.get(), source_);
+  if (allow_precompiled) {
+    const auto precompiled = getCustomSection(getPrecompiledSectionName());
+    if (!precompiled.empty()) {
+      auto vec = wasm::vec<byte_t>::make_uninitialized(precompiled.size());
+      ::memcpy(vec.get(), precompiled.data(), precompiled.size());
+
+      module_ = wasm::Module::deserialize(store_.get(), vec);
+      if (!module_) {
+        // Precompiled module that cannot be loaded is considered a hard error,
```
jplevyak: Log here.  In WAVM we throw a WasmVmException with a description of the problem.  We should standardize on how we want to deal with these sorts of failures.

test/tools/wee8_compile/wee8_compile.cc
```diff
@@ -0,0 +1,185 @@
...
+    return wasm::vec<byte_t>::invalid();
+  }
+
+  const auto module = wasm::Module::make(store.get(), content);
+  if (module == nullptr) {
+    std::cerr << "ERROR: Failed to instantiate WebAssembly module from: " << path << std::endl;
+    return wasm::vec<byte_t>::invalid();
+  }
+
+  // TODO(PiotrSikora): figure out how to hook the completion callback.
```
jplevyak: Please attach issue #

jplevyak:
If you merge from master it will fix the build.

test/tools/wee8_compile/wee8_compile.cc
PiotrSikora: #317

source/extensions/common/wasm/v8/v8.cc
PiotrSikora: WAVM doesn't throw an exception in `load()` (which returns `true/false`, since we're a bit inconsistent with use of exceptions and `bool`).

Furthermore, it doesn't look that it validates that module was successfully loaded, but always returns `true`:
https://github.com/envoyproxy/envoy-wasm/blob/cc4af107e98755ae02e6b2b1e99eb49157c1793b/source/extensions/common/wasm/wavm/wavm.cc#L265-L270

PiotrSikora:
Serialized isolate is (obviously) platform specific... I'll add platform suffix.

PiotrSikora:
@lizan any ideas what happened to the CircleCI jobs?

source/extensions/common/wasm/v8/v8.cc
```diff
@@ -272,39 +313,94 @@ WasmVmPtr V8::clone() {
...
+      pos = section_data_start + section_len;
+    } else {
+      pos += section_len;
     }
-    pos += rest;
   }
   return "";
 }
 
+#if defined(__linux__) && defined(__x86_64__)
```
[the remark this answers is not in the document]
PiotrSikora: We don't have CI to test it, so I'm not comfortable enabling this without having tests proving that it can be loaded across machines.

PiotrSikora:
@jplevyak PTAL

Issue #176, opened 2019-09-03
WASM: Add CEL exec API


mandarjog:
Envoy supports executing CEL expressions. We need an API of type

```
int32 expression_handle  compile_cel_expr(std::string expr);

wasmResult evaluate_cel_expr(handle);
```

Advantages:

1. This way the entire expression is evaluated inside the host. CEL offers safety. Since CEL executes in the host, it will not incur multiple back roundtrips across the VM
2. Expressions can be moved to config, instead of C++ code. This means that when expression changes, we do not need to recompile and deploy WASM code. This saves on transport and v8 cache costs.

PR #332, opened 2019-12-04
context: reuse activation

kyessenov:

- Refactor context to share an activation between invocations. 
The activation is going to be used to evaluate stored expressions in the host per #176 
- Group reflective properties into `wasm` top-level wrapper.

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


kyessenov:
Can I get some attention for this PR?

source/extensions/common/wasm/context.cc
```diff
@@ -1481,15 +1392,18 @@ void Context::log(const Http::HeaderMap* request_headers, const Http::HeaderMap*
...
+  setResponseHeaders(response_headers, response_trailers);
 
   onLog();
 
   access_log_request_headers_ = nullptr;
   // ? request_trailers  ?
   access_log_response_headers_ = nullptr;
   access_log_response_trailers_ = nullptr;
-  access_log_stream_info_ = nullptr;
+  clearRequestHeaders();
```
jplevyak: Clear the stream info  as well.

source/extensions/common/wasm/context.cc
```diff
@@ -1461,10 +1370,12 @@ void Context::onEvent(Network::ConnectionEvent event) {
 void Context::initializeReadFilterCallbacks(Network::ReadFilterCallbacks& callbacks) {
   network_read_filter_callbacks_ = &callbacks;
   network_read_filter_callbacks_->connection().addConnectionCallbacks(*this);
+  setStreamInfo(callbacks.connection().streamInfo());
```
jplevyak: We need to clear the StreamInfo when it is no longer available.  Previously the streaminfo was obtained on every call which required it.  This ensured that we would not have stray pointers.  Even of the streaminfo has a safe lifetime with this change it could change in the future.

source/extensions/common/wasm/context.h
```diff
@@ -57,7 +65,8 @@ class Context : public Logger::Loggable<Logger::Id::wasm>,
                 public Http::StreamFilter,
                 public Network::ConnectionCallbacks,
                 public Network::Filter,
-                public std::enable_shared_from_this<Context> {
+                public std::enable_shared_from_this<Context>,
+                public Filters::Common::Expr::BaseWrapper {
```
jplevyak: Use composition rather than subclassing.  The Context should contain what it needs from CELL rather than being a CELL object.

source/extensions/common/wasm/context.h
```diff
@@ -406,6 +419,80 @@ class Context : public Logger::Loggable<Logger::Id::wasm>,
   uint32_t status_code_{0};
   absl::string_view status_message_;
 
+  Filters::Common::Expr::Activation activation_{};
+
+  // Stream info bound to either a request or a connection.
+  const StreamInfo::StreamInfo* stream_info_{};
+  void setStreamInfo(const StreamInfo::StreamInfo& stream_info) {
```
jplevyak: Move this to the .cc file

source/extensions/common/wasm/context.h
```diff
@@ -406,6 +419,80 @@ class Context : public Logger::Loggable<Logger::Id::wasm>,
...
+    if (&stream_info != stream_info_) {
+      stream_info_ = &stream_info;
+      // note: none of these use an arena to produce a wrapper, which means
+      // they are safe to re-use across evaluations
+      activation_.InsertValueProducer(
+          Filters::Common::Expr::Connection,
+          std::make_unique<Filters::Common::Expr::ConnectionWrapper>(stream_info));
```
jplevyak: Why are we adding the same stream_info five times?

source/extensions/common/wasm/context.h
```diff
@@ -406,6 +419,80 @@ class Context : public Logger::Loggable<Logger::Id::wasm>,
...
+          std::make_unique<Filters::Common::Expr::PeerWrapper>(stream_info, false));
+      activation_.InsertValueProducer(
+          Filters::Common::Expr::Destination,
+          std::make_unique<Filters::Common::Expr::PeerWrapper>(stream_info, true));
+      activation_.InsertValueProducer(
+          Filters::Common::Expr::Metadata,
+          std::make_unique<Filters::Common::Expr::MetadataProducer>(stream_info.dynamicMetadata()));
+    }
+  }
+  void setRequestHeaders(const Http::HeaderMap* request_headers) {
```
jplevyak: Move this code to the .cc file.

source/extensions/common/wasm/context.h
```diff
@@ -406,6 +419,80 @@ class Context : public Logger::Loggable<Logger::Id::wasm>,
...
+  void setResponseHeaders(const Http::HeaderMap* response_headers,
+                          const Http::HeaderMap* response_trailers) {
+    activation_.InsertValueProducer(Filters::Common::Expr::Response,
+                                    std::make_unique<Filters::Common::Expr::ResponseWrapper>(
+                                        response_headers, response_trailers, *stream_info_));
+  }
+  void clearResponseHeaders() { activation_.RemoveValueEntry(Filters::Common::Expr::Response); }
+
+  // An expression wrapper for the WASM state
+  struct WasmStateWrapper : public Filters::Common::Expr::BaseWrapper {
```
jplevyak: Move this entire struct to the .cc file.

test/extensions/filters/http/wasm/wasm_filter_test.cc
```diff
@@ -229,8 +228,7 @@ TEST_P(WasmHttpFilterTest, AccessLog) {
   Buffer::OwnedImpl data("hello");
   EXPECT_EQ(Http::FilterDataStatus::Continue, filter_->decodeData(data, true));
   filter_->onDestroy();
-  StreamInfo::MockStreamInfo log_stream_info;
-  filter_->log(&request_headers, nullptr, nullptr, log_stream_info);
+  filter_->log(&request_headers, nullptr, nullptr, request_stream_info_);
```
jplevyak: Why are we using the request_stream_info for the log() call?  That is a different filter.

source/extensions/common/wasm/context.cc
```diff
@@ -400,20 +375,70 @@ enum class PropertyToken { PROPERTY_TOKENS(_DECLARE) };
 static absl::flat_hash_map<std::string, PropertyToken> property_tokens = {PROPERTY_TOKENS(_PAIR)};
 #undef _PARI
 
-WasmResult Context::getProperty(absl::string_view path, std::string* result) {
-  using google::api::expr::runtime::CelValue;
-  using google::api::expr::runtime::FieldBackedListImpl;
-  using google::api::expr::runtime::FieldBackedMapImpl;
+absl::optional<CelValue> Context::operator[](CelValue key) const {
```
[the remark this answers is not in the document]
kyessenov: Can we discuss whether CEL should be in the ABI to https://github.com/envoyproxy/envoy-wasm/issues/176?

source/extensions/common/wasm/context.cc
kyessenov: I am actually wondering if this is the right place. Is OnLog the last operation before the request/connection is discarded?

source/extensions/common/wasm/context.cc
kyessenov: Agree. Which callback should be clearing the stream info pointer? onHeaders/onBody/etc generally share the same stream info, so it helps with the performance to avoid reconstructing the environment.

source/extensions/common/wasm/context.h
kyessenov: Yeah, I can use the same pattern as WasmStateWrapper.

source/extensions/common/wasm/context.h
kyessenov: These are different wrappers around the same info object. Agree, it's not elegant. I think we would want to have something like getProperty in Activation itself. But that's a change I have to pull through google3, Envoy, etc.

test/extensions/filters/http/wasm/wasm_filter_test.cc
kyessenov: There is one stream info per request / connection. This is an implicit invariant in ABI.

kyessenov:
I'm having second thoughts about this approach. I think I'll just turn Context into an activation, so we just need to implement a single virtual method.

kyessenov:
Superseded by #343

PR #357, opened 2020-01-02
build: support linux-aarch64

jjjighg:
Using native commands gn and ninja, we successfully compiled on aarch64.

Signed-off-by: Dongyuan Tang <tangdongyuan@huawei.com>

Description: In the old file wee8.genrule_cmd, it does not support aarch64, and v8 default gn and ninja do not support aarch64 architecture. Gn and ninja using aarch64 native command can compile successfully.
Risk Level: low
Testing: manual testing
Docs Changes: N/A
Release Notes: N/A
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -27,11 +27,13 @@ build:linux --copt=-fPIC
 # We already have absl in the build, define absl=1 to tell googletest to use absl for backtrace.
 build --define absl=1
 
-# Pass PATH, CC, CXX and LLVM_CONFIG variables from the environment.
+# Pass PATH, CC, CXX, LLVM_CONFIG, GN and NINJA variables from the environment.
 build --action_env=CC
 build --action_env=CXX
 build --action_env=LLVM_CONFIG
 build --action_env=PATH
+build --action_env=GN
+build --action_env=NINJA
 
 # Common flags for sanitizers
 build:sanitizer --define tcmalloc=disabled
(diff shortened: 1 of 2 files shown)


jjjighg:
> The build process was already upstreamed, so this change should be submitted directly to [envoyproxy/envoy](https://github.com/envoyproxy/envoy) and not here. Also, it looks that someone is already working on it upstream ([envoyproxy/envoy#9527](https://github.com/envoyproxy/envoy/pull/9527)) (opened 2 days before your change here), so perhaps you could review it instead?

Thanks for your reply. I think that is also good. It would be better if v8 can support arm64 completely. Should I close this pr?

.bazelrc
```diff
@@ -27,11 +27,13 @@ build:linux --copt=-fPIC
 # We already have absl in the build, define absl=1 to tell googletest to use absl for backtrace.
 build --define absl=1
 
-# Pass PATH, CC, CXX and LLVM_CONFIG variables from the environment.
+# Pass PATH, CC, CXX, LLVM_CONFIG, GN and NINJA variables from the environment.
 build --action_env=CC
 build --action_env=CXX
 build --action_env=LLVM_CONFIG
 build --action_env=PATH
+build --action_env=GN
```
lizan: Don't add extra action envs for each tool, just use what in PATH or buildtools/linux64/gn. If you want override just have them in PATH.

lizan:
And right this should be in upstream, let's close this one.

jjjighg:
ok

PR #361, opened 2020-01-06
wasm: drop absl from time and duration

kyessenov:

We cannot compile absl to Wasm since it has many system calls.

/assign @jplevyak @bianpengyuan

### DIFF
--- a/api/wasm/cpp/proxy_wasm_api.h
+++ b/api/wasm/cpp/proxy_wasm_api.h
@@ -476,7 +476,22 @@ inline Optional<WasmDataPtr> getProperty(std::initializer_list<StringView> parts
   return std::make_unique<WasmData>(value_ptr, value_size);
 }
 
-inline bool getStringValue(std::initializer_list<StringView> parts, std::string* out) {
+// Generic property reader for basic types: int64, uint64, double, bool
+// Durations are represented as int64 nanoseconds.
+// Timetamps are represented as int64 Unix nanoseconds.
+template <typename T>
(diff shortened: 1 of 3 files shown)


api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -476,31 +476,44 @@ inline Optional<WasmDataPtr> getProperty(std::initializer_list<StringView> parts
...
+inline bool getMessageValue(std::initializer_list<StringView> parts,
+                            T* value_ptr) {
   auto buf = getProperty(parts);
-  if (!buf.has_value() || buf.value()->size() != sizeof(T)) {
+  if (!buf.has_value()) {
     return false;
   }
-  *out = *reinterpret_cast<const T*>(buf.value()->data());
-  return true;
+  if (buf.value()->size() == 0) {
```
bianpengyuan: Should this return a pointer to empty T instead of a nullptr? What does ParseFromArray do when size is 0?

kyessenov: Zero array is intentional if you look into the context. nullptr is a valid value that is reported as a zero-size message. ParseFromArray is probably going to return an empty message which is not right, since it's usually implemented via merge.

bianpengyuan:
@jplevyak Can we please merge this? thanks!

jplevyak:
@kyessenov can you merge or do you want me to?

kyessenov:
@jplevyak I don't have merge rights. It's only you, Piotr, and envoy maintainers.

bianpengyuan:
ping, can we merge this?

PR #373, opened 2020-01-14
Add foreign function support to Wasm.

jplevyak:


source/extensions/common/wasm/exports.cc
```diff
@@ -217,6 +217,40 @@ Word done(void* raw_context) {
...
 }
 
+Word call_foreign_function(void* raw_context, Word function_name, Word function_name_size,
+                           Word arguments, Word arguments_size, Word results, Word results_size) {
+  auto context = WASM_CONTEXT(raw_context);
+  auto function = context->wasmVm()->getMemory(function_name.u64_, function_name_size.u64_);
+  if (!function) {
+    return wasmResultToWord(WasmResult::InvalidMemoryAccess);
+  }
+  auto args_opt = context->wasmVm()->getMemory(arguments.u64_, arguments_size.u64_);
```
PiotrSikora: This is really a single argument, not an argument list/vector, right? If so, maybe name it accordingly?

Also, I'm not sure if that matches @kyessenov's requirements.

jplevyak: This is what we negotiated in the design doc.  The two sides have to agree on the structure of the input block and output block.

jplevyak: It is a single "argument", but it is supposed to contain one or more arguments marshalled by the caller.  See https://docs.google.com/document/d/1g3AesTWTD6HKNvEjGj6iTkGV21mhyBZmjygVaCUgjyI/edit?usp=sharing

PR #374, opened 2020-01-15
Add Wasm micro benchmarks.

jplevyak:


test/extensions/wasm/test_data/speed_cpp.cc
```diff
@@ -1,11 +1,97 @@
...
+#ifndef NULL_PLUGIN
 #include "proxy_wasm_intrinsics.h"
+#else
+#include "extensions/common/wasm/null/null_plugin.h"
+#endif
 
-extern "C" PROXY_WASM_KEEPALIVE void proxy_on_tick(uint32_t /* context_zero */) {
+START_WASM_PLUGIN(WasmSpeedCpp)
+
+int xDoNotRemove = 0;
```
PiotrSikora: Why do we need this?

jplevyak: This is to prevent the compiler from eliminating the code.

PR #379, opened 2020-01-24
wasm: add support for wasi_snapshot_preview1.

PiotrSikora:
This is renamed wasi_unstable, so it's trivial to support both.


PiotrSikora:
Note: it looks that we had a bad merge with upstream and ended with very similar (but not the same) test files: `test/extensions/wasm/wasm_test.cc` and `test/extensions/common/wasm/wasm_test.cc`.

This is not a big issue for now, but we should clean this up.

PR #386, opened 2020-01-30
wasm: expose proxy_expr in bazel, fix TSAN warning in ANTLR

kyessenov:

TSAN detected an (innocent) data race in ANTLR parser.
Also, add a build file so we can see it in bazel.

/assign @PiotrSikora 
/assign @jplevyak

### DIFF
--- a/api/wasm/cpp/contrib/BUILD
+++ b/api/wasm/cpp/contrib/BUILD
@@ -1,0 +1,17 @@
+licenses(["notice"])  # Apache 2
+
+load(
+    "//bazel:envoy_build_system.bzl",
+    "envoy_cc_library",
+    "envoy_package",
+)
+
+envoy_package()
+
+envoy_cc_library(
+    name = "contrib_lib",
+    hdrs = [
+        "proxy_expr.h",
+    ],
+)
+
--- a/bazel/antlr.patch
+++ b/bazel/antlr.patch
@@ -24,4 +24,3 @@ index 827c3d59f..62914cf55 100755
    _mode = mode;
    ssize_t mark = input->mark();
 
-d
--- a/bazel/antlr.patch
+++ b/bazel/antlr.patch
@@ -11,4 +11,17 @@ index c6cceda13..e86533759 100755
  }
 
  ATNDeserializer::ATNDeserializer(const ATNDeserializationOptions& dso): deserializationOptions(dso) {
+diff --git a/runtime/Cpp/runtime/src/atn/LexerATNSimulator.cpp b/runtime/Cpp/runtime/src/atn/LexerATNSimulator.cpp
+index 827c3d59f..62914cf55 100755
+--- a/runtime/Cpp/runtime/src/atn/LexerATNSimulator.cpp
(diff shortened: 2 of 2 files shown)


bazel/antlr.patch
```diff
@@ -11,4 +11,17 @@ index c6cceda13..e86533759 100755
...
+@@ -69,7 +69,7 @@ void LexerATNSimulator::copyState(LexerATNSimulator *simulator) {
+ }
+
+ size_t LexerATNSimulator::match(CharStream *input, size_t mode) {
+-  match_calls++;
++  // match_calls++;
+   _mode = mode;
+   ssize_t mark = input->mark();
 
+d
```
PiotrSikora: That seems like a leftover?

kyessenov: oops, sorry. vim and fat fingers.

kyessenov:
PTAL

bazel/antlr.patch
```diff
@@ -11,4 +11,16 @@ index c6cceda13..e86533759 100755
...
+diff --git a/runtime/Cpp/runtime/src/atn/LexerATNSimulator.cpp b/runtime/Cpp/runtime/src/atn/LexerATNSimulator.cpp
+index 827c3d59f..62914cf55 100755
+--- a/runtime/Cpp/runtime/src/atn/LexerATNSimulator.cpp
++++ b/runtime/Cpp/runtime/src/atn/LexerATNSimulator.cpp
+@@ -69,7 +69,7 @@ void LexerATNSimulator::copyState(LexerATNSimulator *simulator) {
+ }
+
+ size_t LexerATNSimulator::match(CharStream *input, size_t mode) {
+-  match_calls++;
++  // match_calls++;
```
PiotrSikora: Could you open issue upstream to get this fixed, so that we won't need to carry this patch forever?

api/wasm/cpp/contrib/BUILD
```diff
@@ -0,0 +1,17 @@
...
+load(
+    "//bazel:envoy_build_system.bzl",
+    "envoy_cc_library",
+    "envoy_package",
+)
+
+envoy_package()
+
+envoy_cc_library(
+    name = "contrib_lib",
```
PiotrSikora: Nit: `expr_lib`, in case we have other features in contrib in the future?

kyessenov: I think we can split this directory once it becomes large. I like rule name matching the directory name.

bazel/antlr.patch
kyessenov: This one is hard. 4.7.2 ANTLR is newer but has more issues (not sure why). I chose 4.7.1 since that's in google3 with similar patches. I'll send a google3 patch with this change.

api/wasm/cpp/contrib/BUILD
PiotrSikora: Sounds good.

bazel/antlr.patch
PiotrSikora: Sounds good, thanks!

PiotrSikora:
@jplevyak please merge if you don't have any objections.

PR #387, opened 2020-01-31
 Cache and share the base Wasm

jplevyak:
Cache and share the base Wasm. Use the new definition of Wasm Key to find the base Wasm and thread local Wasm.

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -69,7 +69,7 @@ std::string Sha256(absl::string_view data) {
 }
 
 std::string Xor(absl::string_view a, absl::string_view b) {
-  ASSERT(a.size == b.size());
+  ASSERT(a.size() == b.size());
   std::string result;
   result.reserve(a.size());
   for (size_t i = 0; i < a.size(); i++) {
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -69,6 +69,7 @@ std::string Sha256(absl::string_view data) {
 }
 
 std::string Xor(absl::string_view a, absl::string_view b) {
+  ASSERT(a.size == b.size());
   std::string result;
   result.reserve(a.size());
   for (size_t i = 0; i < a.size(); i++) {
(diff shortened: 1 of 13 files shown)


source/extensions/common/wasm/wasm.cc
```diff
@@ -65,11 +65,23 @@ std::string base64Sha256(absl::string_view data) {
...
   RELEASE_ASSERT(rc == 1, "Failed to finalize digest");
   EVP_MD_CTX_free(ctx);
-  return Base64::encode(reinterpret_cast<const char*>(&digest[0]), digest.size());
+  return std::string(reinterpret_cast<const char*>(&digest[0]), digest.size());
 }
 
-// Map from Wasm ID to the local Wasm instance.
-thread_local absl::flat_hash_map<std::string, std::weak_ptr<WasmHandle>> local_wasms;
+std::string Xor(absl::string_view a, absl::string_view b) {
+  std::string result;
```
PiotrSikora: Could you add `ASSERT` here to make sure that `a.size() == b.size()`?

source/extensions/common/wasm/wasm.cc
```diff
@@ -243,18 +256,20 @@ void Wasm::getFunctions() {
...
+    : std::enable_shared_from_this<Wasm>(*wasm_handle->wasm()), vm_id_(wasm_handle->wasm()->vm_id_),
+      vm_key_(wasm_handle->wasm()->vm_key_),
+      started_from_(wasm_handle->wasm()->wasm_vm()->cloneable()),
+      scope_(wasm_handle->wasm()->scope_), cluster_manager_(wasm_handle->wasm()->cluster_manager_),
+      dispatcher_(dispatcher), time_source_(dispatcher.timeSource()),
+      base_wasm_handle_(wasm_handle), wasm_stats_(wasm_handle->wasm()->wasm_stats_),
+      stat_name_set_(wasm_handle->wasm()->stat_name_set_) {
   if (started_from_ != Cloneable::NotCloneable) {
-    wasm_vm_ = wasm.wasm_vm()->clone();
+    wasm_vm_ = wasm_handle->wasm()->wasm_vm()->clone();
```
PiotrSikora: Nit: Use `base_wasm_handle_` instead of `wasm_handle` (I think it's better for readability, but feel free to ignore this).

source/extensions/common/wasm/wasm.cc
```diff
@@ -495,13 +502,57 @@ static void createWasmInternal(const VmConfig& vm_config, PluginSharedPtr plugin
                  .value_or(code.empty() ? EMPTY_STRING : INLINE_STRING);
   }
 
-  auto callback = [wasm, plugin, cb, source, allow_precompiled = vm_config.allow_precompiled(),
+  auto callback = [vm_config, scope, &cluster_manager, &dispatcher, plugin, cb, source,
+                   allow_precompiled = vm_config.allow_precompiled(),
```
PiotrSikora: `vm_config` is now passed in, so you probably don't need to pass `allow_precompiled` separately.

source/extensions/common/wasm/wasm.cc
```diff
@@ -495,13 +502,57 @@ static void createWasmInternal(const VmConfig& vm_config, PluginSharedPtr plugin
...
     std::unique_ptr<Context> context(context_ptr);
     if (code.empty()) {
       throw WasmException(fmt::format("Failed to load WASM code from {}", source));
     }
+    // Construct a unique identifier for the VM based on a hash of the code, vm configuration data
+    // and vm_id.
+    std::string vm_key = Sha256(vm_config.vm_id());
+    vm_key = Xor(vm_key, Sha256(vm_config.configuration()));
+    vm_key = Xor(vm_key, Sha256(code));
+    vm_key = Base64::encode(&*vm_key.begin(), vm_key.size());
```
PiotrSikora: Is this done to avoid creating a temporary string?

source/extensions/common/wasm/wasm.cc
```diff
@@ -495,13 +502,57 @@ static void createWasmInternal(const VmConfig& vm_config, PluginSharedPtr plugin
...
+      absl::MutexLock l(&base_wasms_mutex_);
+      auto it = base_wasms_->find(vm_key);
+      if (it != base_wasms_->end()) {
+        wasm = it->second.lock();
+        if (!wasm) {
+          base_wasms_->erase(it);
+        }
+      }
+    }
+    // Drop the lock so that the the expensive Wasm creation is not under the lock.
```
PiotrSikora: But then, we will pay the price for creating many expensive Wasm VMs at the same time, and discarding those that lost the race anyway, won't we?

source/extensions/common/wasm/wasm.cc

source/extensions/common/wasm/wasm.cc

source/extensions/common/wasm/wasm.cc
jplevyak: Yes, 'code' can be large.   However, generally if you concatenate then you have to reason about separaters and whether or not bits of one string are legal prefixes of others.  Folding the SHA together with XOR avoids that.

source/extensions/common/wasm/wasm.cc
jplevyak: Given that this is going to run on the main thread currently there are not going to be any races. So, perhaps it isn't worth dropping the lock.  Removed.  PTAL

source/extensions/common/wasm/wasm.cc
```diff
@@ -495,22 +505,62 @@ static void createWasmInternal(const VmConfig& vm_config, PluginSharedPtr plugin
...
+        if (!wasm) {
+          base_wasms_->erase(it);
+        }
+      }
+      if (!wasm) {
+        wasm = std::make_shared<WasmHandle>(std::make_shared<Wasm>(
+            vm_config.runtime(), vm_config.vm_id(), vm_config.configuration(), vm_key, scope,
+            cluster_manager, dispatcher));
+        {
+          auto it = base_wasms_->find(vm_key);
```
PiotrSikora: Do we still need this? This should be all happening under the lock, so nothing can be added there, right?

source/extensions/common/wasm/wasm.cc
```diff
@@ -495,22 +505,50 @@ static void createWasmInternal(const VmConfig& vm_config, PluginSharedPtr plugin
...
+      if (it != base_wasms_->end()) {
+        wasm = it->second.lock();
+        if (!wasm) {
+          base_wasms_->erase(it);
+        }
+      }
+      if (!wasm) {
+        wasm = std::make_shared<WasmHandle>(std::make_shared<Wasm>(
+            vm_config.runtime(), vm_config.vm_id(), vm_config.configuration(), vm_key, scope,
+            cluster_manager, dispatcher));
```
PiotrSikora: Can this ever return `nullptr`? Do we need to handle it or are you following the "fail/crash-fast" principles here?

jplevyak: make_shared is never going to return nullptr, and if we OOM, well, we fail fast.

PR #390, opened 2020-01-31
Merge envoyproxy/master.

jplevyak:


PiotrSikora:
This is bad merge, it removes #382, which is why CI is broken.

jplevyak:
I have rerun the ASAN 8 times now.   This is unreasonable toil.   What can we do to fix it? 


ERROR: /source/test/mocks/server/BUILD:11:1: C++ compilation of rule '//test/mocks/server:server_mocks' failed (Exit 254) clang failed: error executing command 
  (cd /build/tmp/_bazel_bazel/b570b5ccd0454dc9af9f65ab1833764d/sandbox/processwrapper-sandbox/2242/execroot/envoy && \

jplevyak:
clang-9: error: unable to execute command: Killed

PiotrSikora:
@jplevyak see my comment above!

#345 broke build (which `asan` correctly signaled), because CEL/ANTLR dependency isn't available in `//http-filter-example` repo that tests integration with other workspaces, but it was merged anyway...

I fixed CI by eliminating that test from the `asan` builds in #382 (since we don't really care about it for the purpose Wasm development), but you incorrectly removed in this merge, so `asan` is broken again (legitimately, not flaky). Rerunning it won't help, but if you restore #382, then it will pass.

PR #394, opened 2020-02-05
Add access to upstreamfilterstate in WasmStateWrapper

gargnupur:
Add access to upstreamfilterstate in WasmStateWrapper 

Description: 
Risk Level:
Ref:  https://github.com/istio/istio/issues/20802
Testing: Tested E2E in istio/proxy
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/tcp_proxy/tcp_proxy.cc
+++ b/source/common/tcp_proxy/tcp_proxy.cc
@@ -239,6 +239,12 @@ void Filter::initialize(Network::ReadFilterCallbacks& callbacks, bool set_connec
   getStreamInfo().setDownstreamLocalAddress(read_callbacks_->connection().localAddress());
   getStreamInfo().setDownstreamRemoteAddress(read_callbacks_->connection().remoteAddress());
   getStreamInfo().setDownstreamSslConnection(read_callbacks_->connection().ssl());
+  read_callbacks_->connection().streamInfo().setDownstreamLocalAddress(
+      read_callbacks_->connection().localAddress());
+  read_callbacks_->connection().streamInfo().setDownstreamRemoteAddress(
+      read_callbacks_->connection().remoteAddress());
+  read_callbacks_->connection().streamInfo().setDownstreamSslConnection(
+      read_callbacks_->connection().ssl());
 
   // Need to disable reads so that we don't write to an upstream that might fail
   // in onData(). This will get re-enabled when the upstream connection is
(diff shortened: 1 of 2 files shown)


source/extensions/common/wasm/context.cc
```diff
@@ -348,32 +348,29 @@ WasmResult serializeValue(Filters::Common::Expr::CelValue value, std::string* re
...
-          const WasmState& result = connection_filter_state_->getDataReadOnly<WasmState>(value);
-          return google::api::expr::runtime::CelValue::CreateBytes(&result.value());
-        }
-      } catch (const EnvoyException& e) {
-        return {};
-      }
-      return {};
     }
+
+    if (upstream_connection_filter_state_ &&
```
kyessenov: What was the outcome of the discussion for the fallback of the filter state to upstream connection filter state?

gargnupur: Don't think we discussed that...
https://github.com/envoyproxy/envoy/pull/9202/files PR added sustaining filterstate to downstream connection lifespan and then I added this https://github.com/envoyproxy/envoy/pull/9896 to share filterstate between upstream and downstream filters.
I see your point that we will suffer from the problem, what if upstream connection closes before downstream connection, then in that case upstream filterstate will be invalid. Is it possible for that to happen?

kyessenov: It's a shared_ptr so the state will outlive the upstream connection. But we should be careful to clear it in the wasm context eventually.

kyessenov: My question was about filter_state itself falling through to upstream connection filter state. E.g. make it try HTTP, downstream TCP, then upstream TCP.

gargnupur: No, didnt try to make filter_state itself fall through to upstream TCP

gargnupur: Talked offline.. will simplify in a different pr..

source/common/tcp_proxy/tcp_proxy.cc
```diff
@@ -239,6 +239,12 @@ void Filter::initialize(Network::ReadFilterCallbacks& callbacks, bool set_connec
   getStreamInfo().setDownstreamLocalAddress(read_callbacks_->connection().localAddress());
   getStreamInfo().setDownstreamRemoteAddress(read_callbacks_->connection().remoteAddress());
   getStreamInfo().setDownstreamSslConnection(read_callbacks_->connection().ssl());
+  read_callbacks_->connection().streamInfo().setDownstreamLocalAddress(
```
jplevyak: Is this being upstreamed?

gargnupur: I will after getting feedback from our team...

gargnupur: @jplevyak : John removed tcp proxy change her and added to envoy only first, will merge here after that... https://github.com/envoyproxy/envoy/pull/9949

Can you please help merge this PR?

PR #398, opened 2020-02-05
Add tests for multiple root_ids.

jplevyak:


api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -431,6 +431,7 @@ using ContextFactory = std::function<std::unique_ptr<Context>(uint32_t id, RootC
 struct RegisterContextFactory {
   explicit RegisterContextFactory(ContextFactory context_factory,
                                   RootFactory root_factory = nullptr, StringView root_id = "");
+  explicit RegisterContextFactory(ContextFactory root_factory, StringView root_id);
```
PiotrSikora: `s/root_factory/context_factory/`

api/wasm/cpp/proxy_wasm_api.h
```diff
@@ -431,6 +431,7 @@ using ContextFactory = std::function<std::unique_ptr<Context>(uint32_t id, RootC
...
   explicit RegisterContextFactory(RootFactory root_factory, StringView root_id = "");
```
PiotrSikora: Those 3 variants together are extremely confusing. Why can't we pass `nullptr` to `context_factory` or `root_factory` when needed and always use the version with 3 parameters? They all seem to do the same.

api/wasm/cpp/proxy_wasm_api.h
jplevyak: Done.

api/wasm/cpp/proxy_wasm_api.h
jplevyak: Fixed it using a single constructor and two wrappers.  PTAL.

PR #384, opened 2020-01-28
wasm: API to add tags to active span from wasm filter

mohit-a21:
Signed-off-by: Mohit Agarwal <mohit@traceable.ai>

Description: This change adds an API, `activeSpanSetTag` which allows a filter to customize span info and help in adding tracing info.
Risk Level: Low
Testing: As part of the change, test cases have been added along the lines of test cases for existing APIs. Along with that manual testing has been done to verify the API.
Docs Changes: Docs have been updated regarding the API so added.
Fixes: #383

PR #351, opened 2019-12-19
stub out rust api wasm proxy callbacks

joshua-rutherford:
I'm looking to extend the Rust SDK but wanted to get these stubs in as I'm not sure the next time I'll be back here.  I am new to Rust so feeling my way out here a bit.  I've set up my tool chain to build with wasm-pack as the README.md was sparse on configuring the environment.  Is there some other tool chain recommended?

PR #399, opened 2020-02-06
Context support for Rust wasm filter SDK

Shikugawa:
Description: Context support for Rust wasm filter SDK. This is the only proposal implementation for SDK. So It may be required to discuss.
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


api/wasm/rust/src/context.rs
```diff
@@ -0,0 +1,142 @@
...
+  fn on_create(&self);
+}
+
+pub trait RootContextFactory {
+  fn create(&self) -> Arc<Mutex<dyn RootContext + Sync + Send>>;
+}
+pub trait ContextFactory {
+  fn create(
+    &self,
+    _root_context: Arc<Mutex<dyn RootContext + Sync + Send>>,
```
yuval-k: i dont think you need Arc \ mutex here.
regular Rc without mutex should work, as IIRC VMs are instantiated per thread.
even if that wasn't the case i would still go for Arc but without mutex (similar to how envoy filters work today).

Shikugawa: It may be not required Mutex, but required Arc. Because we use only Rc, we can't guarantee the thread safety of static region variables and rustc can't pass them. It can pass with `thread_local!` macro. But we can't apply `thread_local!` and `lazy_static!` simultaneously.

So we should rewrite it to 
```
Arc<dyn RootContext + Sync + Send>
```

Shikugawa:
\restart

Shikugawa:
\retest

PiotrSikora:
Hi @Shikugawa, thanks writing this code and apologies for the delay!

However, please see: https://github.com/envoyproxy/envoy-wasm/pull/351#issuecomment-568057579 and https://github.com/envoyproxy/envoy-wasm/pull/384#issuecomment-580887721. I have a fairly more complete Rust SDK staged for publishing, but it's waiting for some approvals. Once it's out, I'm happy to incorporate changes from this PR. Sorry for the inconvenience!

Shikugawa:
@PiotrSikora Thanks. I'm developing Rust support not only context but also other functionalities. (e.g. pair marshaling/unmarshaling) I will add some functionality if there is missing that. I will help Rust SDK development!

PiotrSikora:
As you already noticed, the Rust SDK has been published at:
https://github.com/proxy-wasm/proxy-wasm-rust-sdk

Please feel free to reopen PRs there, thanks!

PR #401, opened 2020-02-06
Return the foreign function result.

jplevyak:

### DIFF
--- a/source/extensions/common/wasm/exports.cc
+++ b/source/extensions/common/wasm/exports.cc
@@ -237,7 +237,7 @@ Word call_foreign_function(void* raw_context, Word function_name, Word function_
   uint64_t address = 0;
   void* result = nullptr;
   size_t result_size = 0;
-  f(wasm, args, [&wasm, &address, &result, &result_size](size_t s) -> void* {
+  auto res = f(wasm, args, [&wasm, &address, &result, &result_size](size_t s) -> void* {
     result = wasm.allocMemory(s, &address);
     result_size = s;
     return result;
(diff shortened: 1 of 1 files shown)


mandarjog:
needs to be cherry-picked to 1.5

jplevyak:
Responding to Piotr.   The foreign function API has provision for returning an error.  If there is an error which fits in the WasmResult semantics, I think it is reasonable to include it instead of the OK.   I agree that it does change the semantics, but it is not unreasonable.   Let us discuss this on the ABI document and I will merge this and if we have to change it as part of upstreaming or the ABI document review we can do it then.

source/extensions/common/wasm/exports.cc
```diff
@@ -248,7 +248,7 @@ Word call_foreign_function(void* raw_context, Word function_name, Word function_
   if (!context->wasmVm()->setWord(results_size.u64_, Word(result_size))) {
     return wasmResultToWord(WasmResult::InvalidMemoryAccess);
   }
```
PiotrSikora: If `res != WasmResult::Ok` then you probably shouldn't be writing the `result` above.

PiotrSikora: @jplevyak could you fix that? Since this is already merged.

jplevyak: Perhaps.  Again.  I will defer to reviewers.  I am OK providing more information to the caller assuming that the foreign function provided it as this that is part of a private API between the caller and the callee.

kyessenov: It's fine to pass error details.

PiotrSikora: Usually, when call fails (i.e. `res != WasmResult::Ok`), it means that results were not necessarily initialized, and `result` and `result_size` may point to garbage.

PR #404, opened 2020-02-10
wasm: remove unused "free" function.

PiotrSikora:


kyessenov:
Will this break older Wasm modules in any way?

kyessenov:
Also, I don't understand how we can maintain a stable memory footprint in a long running module without `free`?

PiotrSikora:
We immediately pass the memory ownership into VM after requesting it and copying data, so host never holds to any VM-allocated memory.

No, it won't break old modules.

kyessenov:
Ack.

kyessenov:
Does that mean that Wasm holds onto host memory after a memory pressure event? E.g. if module allocates 1GB it cannot give it back to the host?

PiotrSikora:
> Does that mean that Wasm holds onto host memory after a memory pressure event? E.g. if module allocates 1GB it cannot give it back to the host?

Wasm VM allocates the whole linear memory block ("RAM") upfront, but it can grow it as needed (up to configureable maximum). I'm not sure if it can actually shrink it back.

kyessenov:
Yeah, but it sounds like we need some memory management capability to restart modules and revoke memory. Otherwise, it's prone to DoS. I can imagine some of our modules allocate based on the data path input.

PiotrSikora:
Yup.

PR #409, opened 2020-02-13
proxy_set_header_map_pairs doesn't add headers to the request/response correctly

mohit-a21:
Description:
From the docs:
```
Sets request headers with the given header pairs. For each header key-value pair, it acts the same way as replaceRequestHeader.
```
Currently, proxy_set_header_map_pairs is not exporting the pairs to be set and
setHeaderMapPairs is deleting all the existing headers. This pr fixes export of
pairs and replacing of headers in the map

Signed-off-by: Mohit Agarwal <mohit@traceable.ai>

Risk Level: Debateable
Testing: Added test cases
Docs Changes: Not needed.


mohit-a21:
I am not sure if this change is needed or not. Please correct me if I misunderstood the `proxy_set_header_map_pairs ` API.

PiotrSikora:
Hi @mohit-a21,
apologies for the late response. I'm looking at this PR and the existing code, and I believe that you're misinterpreting this API (or I am). Also, the docs might be wrong.

The purpose of this function is to set/replace **complete** header map, not to update only headers passed to this function.

cc @jplevyak

PR #411, opened 2020-02-13
Update GrpcService to add StsService.

jplevyak:


jplevyak:
Once this is in it should be backported... .to 1.5 + 1.4 ?

PiotrSikora:
@duderino @howardjohn FYI

kyessenov:
Definitely 1.5. Desirable in 1.4.

duderino:
Why 1.4?  1.4 is for patches not features.  We violate that sometimes, but it needs a good reason

kyessenov:
So there are two protos for GrpcService. We updated one in 1.4 with StsService field but not the other. Wasm ABI converts the two protos at runtime, so it's losing the field on the ABI.

duderino:
@kyessenov,  @howardjohn and I chatted and we're ok with the change itself.  the concern is more about timing.  If it had arrived a day earlier we could have gotten it into 1.4.5.  I think we can backport this to 1.4, but I don't thing we want to do a special release just for this

Issue #414, opened 2020-02-17
gRPC CreateInitialMetadata callback not work


bianpengyuan:
onCreateInitialMetadata is triggered inline in SendRaw, at that time grpc call token is not yet returned into module, which makes onCreateInitialMetadata always fail. This is the tentative fix in istio envoy repo: https://github.com/istio/envoy/pull/152/files#diff-3367fe470b910c37aac0a6fd5856a936 not sure if it is the proper fix. If yes I could patch the same fix in envoy-wasm repo.

PiotrSikora:
Looking at this now, I'm wondering if there is any reason to have this callback at all? It seems that we're mimicking Envoy's APIs here, but `proxy_grpc_call` could be simply include initial metadata as one of the input parameters, right? or am I missing some use case?

kyessenov:
+1. Metadata should just be part of the call.

bianpengyuan:
How do you express other operation other than adding if initial metadata becomes a param?

PiotrSikora:
> How do you express other operation other than adding if initial metadata becomes a param?

Do you have a real use case for that?

bianpengyuan:
No I don't. This is the first time I tried out this callback :) . I am fine with eliminating it from the ABI and turn it into a param. Will update the PR to do that.

kyessenov:
No ABI change in 1.4 please. Ideally, 1.5 either. The ABI cleanup needs to happen after the release.

bianpengyuan:
No more change to 1.4 for sure. My current priority is to get something working for release-1.5, no matter it involves an ABI change or not.

@PiotrSikora Would it be acceptable if we just do the minimal fix in release-1.5 and master like release-1.4 (https://github.com/istio/envoy/commit/37dbbd43ed8020c4066e0737d1fc7902fceef36a#diff-3367fe470b910c37aac0a6fd5856a936), which involves no abi change. I will also add a test for it. Later on we can clean up the ABI after release-1.5 is out.

PiotrSikora:
I'm fine with having the minimal fix in `release-1.4` and `release-1.5`, but I'd rather do a proper fix in `master`.

bianpengyuan:
Ok then I will do a cherry-pick of release-1.4 change to release-1.5 with a test. Then update master later to eliminate the callback.

jplevyak:
Fixed

PR #417, opened 2020-02-19
Fix Wasm gRPC initial metadata callback

bianpengyuan:
https://github.com/envoyproxy/envoy-wasm/issues/414

This change adds two member variables to gRPC handler, so that onCreateInitialMetadata could be triggered even the call token is not ready in the module.

cc @PiotrSikora 

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


PiotrSikora:
The test failures are real, you need to update API, regenerate SDK and/or update test files.

Also, could you please add a test exercising gRPC callout (I thought we had one, so I'm not sure why it didn't catch this issue?), to make sure that the fix works?

bianpengyuan:
Yeah, wasm gen make file misses root_id.wasm. Seems like gRPC test does not exercise initial metadata code path. Will add a case for it.

bianpengyuan:
Supersede by https://github.com/envoyproxy/envoy-wasm/pull/461

PR #420, opened 2020-02-21
Fix bad merge.

PiotrSikora:
While there, migrate Wasm access logger to the new-style name,
and switch to the REGISTER_FACTORY macro.


PiotrSikora:
cc @gargnupur

PR #421, opened 2020-02-21
ci: test all targets to avoid breaking build.

PiotrSikora:


PiotrSikora:
cc @gargnupur

PR #427, opened 2020-02-25
Fix use-after-free error for remote .wasm file accesses.

jplevyak:


mandarjog:
Can we add a test for remote load? 
Remote load is a fine option to test out functionality.

jplevyak:
Is there any way of mocking the remote data fetcher? There doesn't seem to be.

jplevyak:
The RemtoteAsyncDataProviderPtr is *returned* by the code which uses it instead of being an argument which can be mocked like all the other envoy apis (elements of the Api class).

Unless I am missing something, getting a test in for this is going to require a complete rewrite of the underlying RemoteAsyncDataProvider framework.

kyessenov:
I think we need a test since the feature clearly regressed at some point.
We can whatever setup is best. Feel free to use any of the N test frameworks we got in istio.

jplevyak:
This is going to be in Envoy proper, so the test should be there.

PiotrSikora:
I hate to do this, but could we verify that this PR fixes the issue manually (@mandarjog) and merge this in now in order to get this into 1.5? It doesn't sound like the test is trivial to write, and we could do it in a follow-up.

What do you think?

jplevyak:
Added a test.

PTAL

PR #429, opened 2020-02-25
Make callout responses accessible from stream context.

PiotrSikora:


PiotrSikora:
Same is probably true for the gRPC calls, let me know if I should apply the same fix there, or if you want to fix it differently... or if you don't agree that's an issue in the first place.

source/extensions/common/wasm/context.cc
```diff
@@ -648,6 +648,8 @@ Http::HeaderMap* Context::getMap(HeaderMapType type) {
   }
 }
 
+#define GET_FROM_ROOT(x) isRootContext() ? x : root_context_->x
```
jplevyak: You could just make this an inline function.  Up2U

PiotrSikora: Done, PTAL.

Issue #425, opened 2020-02-24
Support fd_read and fd_fdstat_get in WASI interface


gbrail:
For some code I am generating using C++ and Emscripten, it is very hard to prevent emscripten from adding imports of the "fd_read" and "fd_fdstat_get" functions. I would like to add them to the WASI implementation in Envoy.

I already have a PR and these functions work:

fd_read always returns "EBADF"
fd_fdstat_get returns "EBADF" for anything other than fds 1 and 2 -- since we already allow code to write to those fds using fd_write, the implementation returns a reasonable response for those.

Before I submit a PR, I'd like to know what a reasonable testing strategy is. I'd be willing to attempt a stand-alone unit test for the WASI support in Envoy. Would that be helpful?

PR #431, opened 2020-02-25
wasm: Add new functions to the WASI interface.

gbrail:
Description:

Add some additional symbols to the WASI interface to support
methods that are often imported by Emscripten-generated code:

* fd_read: Always returns EBADF
* fd_fdstat_get: Returns a reasonable response for fds 0 and 1, and EBADF otherwise.

Testing:

Add a "WASI" test configuration to the existing tests that checks
for proper implementation of these by testing "fprintf,"
"fread," and "isatty".

Fixes: https://github.com/envoyproxy/envoy-wasm/issues/425


test/extensions/common/wasm/wasm_test.cc
```diff
@@ -354,6 +354,47 @@ TEST_P(WasmCommonTest, Foreign) {
   wasm->startForTesting(std::move(context), plugin);
 }
 
+TEST_P(WasmCommonTest, WASI) {
+  if (GetParam() != "v8") {
```
PiotrSikora: `if (GetParam() == "null")`, since we support other runtimes.

source/extensions/common/wasm/exports.cc
```diff
@@ -710,6 +717,26 @@ Word wasi_unstable_fd_seek(void*, Word, int64_t, Word, Word) {
 // __wasi_errno_t __wasi_fd_close(__wasi_fd_t fd);
 Word wasi_unstable_fd_close(void*, Word) { throw WasmException("wasi_unstable fd_close"); }
 
+// __wasi_errno_t __wasi_fd_fdstat_get(__wasi_fd_t fd, __wasi_fdstat_t *stat)
+Word wasi_unstable_fd_fdstat_get(void* raw_context, Word fd, Word statOut) {
+  // We will only support this interface on stdout and stderr
+  if ((fd.u64_ < 1) || (fd.u64_ > 2)) {
```
PiotrSikora: `if (fd.u64_ != 1 && fd.u64_ != 2) { ... }`

source/extensions/common/wasm/exports.cc
```diff
@@ -701,6 +701,13 @@ Word wasi_unstable_fd_write(void* raw_context, Word fd, Word iovs, Word iovs_len
   return 0; // __WASI_ESUCCESS
 }
 
+// __wasi_errno_t __wasi_fd_read(_wasi_fd_t fd, const __wasi_iovec_t *iovs,
+//    size_t iovs_len, __wasi_size_t *nread);
+Word wasi_unstable_fd_read(void*, Word, Word, Word, Word) {
+  // Don't support reading of any files.
+  return 8; // __WASI_EBADF
```
PiotrSikora: What about using `NOSYS` (Function not supported) or `NOTCAPABLE` (Extension: Capabilities insufficient)?

gbrail: ENOSYS works -- interestingly there's a different value in the Emscripten C code than on my Linux system but it works. ENOTCAPABLE doesn't seem to be defined at least in my Emscripten yet.

gbrail: And if that makes sense perhaps in another PR we can update some other existing APIs the same way.

PiotrSikora: Let's go with `ENOSYS` for clearly unimplemented functions, and it definitely makes sense to update other functions as well. Thanks!

cc @jplevyak

source/extensions/common/wasm/exports.cc
gbrail: Fixed in the next commit in this PR.

test/extensions/common/wasm/wasm_test.cc
gbrail: Fixed in the next commit in this PR.

source/extensions/common/wasm/exports.cc
```diff
@@ -710,6 +717,26 @@ Word wasi_unstable_fd_seek(void*, Word, int64_t, Word, Word) {
 // __wasi_errno_t __wasi_fd_close(__wasi_fd_t fd);
 Word wasi_unstable_fd_close(void*, Word) { throw WasmException("wasi_unstable fd_close"); }
 
+// __wasi_errno_t __wasi_fd_fdstat_get(__wasi_fd_t fd, __wasi_fdstat_t *stat)
+Word wasi_unstable_fd_fdstat_get(void* raw_context, Word fd, Word statOut) {
+  // We will only support this interface on stdout and stderr
+  if ((fd.u64_ != 1) && (fd.u64_ != 2)) {
```
PiotrSikora: Style: you're overusing parentheses here and in tests.

gbrail: Done!

PR #432, opened 2020-02-25
Delay sendLocalReply till after the call to prevent rentrant calls.

jplevyak:

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1103,10 +1103,10 @@ bool Context::validateConfiguration(absl::string_view configuration, PluginShare
 }
 
 bool Context::onConfigure(absl::string_view plugin_configuration, PluginSharedPtr plugin) {
-  DeferAfterCallActions actions(this);
   if (!wasm_->on_configure_) {
     return true;
   }
+  DeferAfterCallActions actions(this);
   configuration_ = plugin_configuration;
   plugin_ = plugin;
   auto result =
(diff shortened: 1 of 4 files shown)


source/extensions/common/wasm/context.cc
```diff
@@ -1091,6 +1103,7 @@ bool Context::validateConfiguration(absl::string_view configuration, PluginShare
 }
 
 bool Context::onConfigure(absl::string_view plugin_configuration, PluginSharedPtr plugin) {
+  DeferAfterCallActions actions(this);
   if (!wasm_->on_configure_) {
     return true;
   }
```
PiotrSikora: Those could be moved to after the check if the callback actually exist, right? Otherwise, we're wasting cycles.

(Same in most other places)

source/extensions/common/wasm/wasm.h
```diff
@@ -141,6 +141,14 @@ class Wasm : public Logger::Loggable<Logger::Id::wasm>, public std::enable_share
     return true;
   }
 
+  void addAfterVmCallAction(std::function<void()> f) { after_vm_call_actions_.push_back(f); }
```
PiotrSikora: You're both adding at and removing from the back of the queue, which makes this LIFO. I think it should be FIFO instead.

source/extensions/common/wasm/wasm.h
```diff
@@ -141,6 +141,14 @@ class Wasm : public Logger::Loggable<Logger::Id::wasm>, public std::enable_share
...
+  void doAfterVmCallActions() {
+    while (!after_vm_call_actions_.empty()) {
+      after_vm_call_actions_.back()();
+      after_vm_call_actions_.pop_back();
+    }
```
PiotrSikora: This gets recursive and the deferred functions are called indefinitely before this code has a chance to remove them from the queue. I'm not sure why, to be honest.

One way to work around this is to make the call only after removing the function from the queue, i.e.
```
auto f = after_vm_call_actions_.back();
after_vm_call_actions_.pop_back();
f();
```
but that probably only works if the is only one deferred call?

It should be fine  in the loop so long as the function to be called is removed first as you suggest.

source/extensions/common/wasm/context.cc
jplevyak: True, but it was more obviously correct, but Done.

source/extensions/common/wasm/wasm.h

PR #433, opened 2020-02-26
ci: increase timeout for a few //test/common/stats:... tests.

PiotrSikora:
Those tests occasionally timeout on istio/envoy CI.


PiotrSikora:
@jplevyak This is mostly to future-proof import into `envoy/istio`, since those tests are timing out there. Added note to the PR description.

PR #434, opened 2020-02-26
Add missing deferred operation handlers.

jplevyak:


PiotrSikora:
Force-merging, since ASan fails because of #435.

PR #437, opened 2020-02-27
wasm: Add API to get buffered body data.

gbrail:
Description: Change the behavior of "proxy_on_request_body" and "proxy_on_response_body" to make it easy to get accessed to the buffered body.

When one of these callbacks returns "StopIterationAndBuffer," the next invocation of that callback will return the data that Envoy has been buffering. Specifically, the "length" parameter passed to the callback will reflect the buffered data, and the various "getBuffer" methods will return the data buffered so far.

A filter that wishes to gather the whole body in one buffer and act on it then can return StopIterationAndBuffer from each call until the "end_of_stream" flag is set, and then it can access the entire body at once.

A filter that wishes to only examine part of the body can return StopIterationAndBuffer until the buffer contains enough data, then return another status code to stop buffering.

Like in other Envoy plugins, Envoy will not allow the buffer to exceed the currently configured maximum. If this happens, Envoy will automatically stop filter processing and return either 413 (Payload Too Large) on the request path, or 500 (Internal Server Error) on the response path.

Testing: New test cases added to wasm_filter_test.cc, with a new WASM module to test it.


api/wasm/cpp/proxy_wasm_common.h
```diff
@@ -73,7 +73,8 @@ enum class BufferType : int32_t {
   NetworkUpstreamData = 3,   // During the onLog callback these are immutable
   HttpCallResponseBody = 4,  // Immutable
   GrpcReceiveBuffer = 5,     // Immutable
-  MAX = 5,
+  BufferedBody = 6,          // After returning StopIterationAndBuffer
```
PiotrSikora: I'm strongly against introducing another enum for this purpose.

Why this buffer cannot be returned using `getBuffer(HttpRequestBody)` and `getBuffer(HttpResponseBody)`?

gbrail: If a filter returns StopIterationAndBuffer from encode/decodeData, then Envoy gathers all the data in each chunk into a single buffer. It also enforces a maximum buffer length and rejects the request or fails the response if it's too big. That buffer is available by calling "decodingBuffer" or "encodingBuffer" on the filter callbacks, and this PR makes it possible to call those functions. (There's another function used to modify it, and we'll get to that later.) From what I can tell, the Lua and DynamoDb filters seem to use this capability.

Without a way to access this buffer, then a filter that wants to look at all the body data at once has to buffer the data itself by aggregating all buffers passed to encode/decodeData, and handle its own logic to prevent running out of memory. Or, the filter writer has to tell users to always include the special "buffer" filter before their filter.

gbrail: Also -- if you want to have a longer conversation...

If one is trying to write a filter that validates the message body and either rejects or accepts it (for instance, it wants to parse some JSON) or if it wants to read a message body and then write out a new one (for instance, it wants to then transform that JSON) then we need some way for the filter to safely and efficiently buffer the body and change it later.

However, unless people are going to use the WASM support to build lower-level filters to do things like compression, then we don't need all the complexity of the full Envoy API -- we could instead design something with a  simpler interface. I have some ideas on that and I'd be happy to propose one if you'd like, but I assumed that the goal if this project was to offer the same flexibility to WASM authors that Envoy offers C++ programmers.

PiotrSikora: @gbrail sorry if my response wasn't clear, I definitely want to have access to the buffered data for obvious reasons. What I'm suggesting is that this data should be available via `getBuffer(HttpRequestBody)` call (i.e. consumer shouldn't care whether this is new or buffered data) and not via a separate call. Does it make sense?

gbrail:
Definitely... let me make sure I can think of an interface that works,
though, because we might have to keep track of state in the "context"
object.

Maybe proxy_get_buffer_*(HttpRequestBody) works like this:

1) If the filter previously returned StopIterationAndBuffer, then return
the buffered data
2) Otherwise, return the current chunk.

That is not how regular callbacks work but IMO it is indeed simpler to use,
so I can try and test that out if you agree.

source/extensions/common/wasm/context.cc
```diff
@@ -1181,17 +1187,23 @@ Http::FilterHeadersStatus Context::onRequestHeaders() {
...
 }
 
-Http::FilterDataStatus Context::onRequestBody(int body_buffer_length, bool end_of_stream) {
+Http::FilterDataStatus Context::onRequestBody(int, bool end_of_stream) {
   if (!wasm_->on_request_body_) {
     return Http::FilterDataStatus::Continue;
   }
+  if (!end_of_stream_ && buffering_body_) {
+    return Http::FilterDataStatus::StopIterationAndBuffer;
+  }
```
PiotrSikora: This forces consumers to select between streaming or waiting for the complete response.

However, this won't work, because large body (i.e. larger than the configured buffer limit) will get pushback, and the complete response will never arrive.

Also, there is a lot of use cases, when the consumer will want to wait until there are X bytes available (e.g. to process `<head>` section of the HTML body, or to get `PROXY-PROTOCOL` header in TCP, etc.), but not necessarily until the complete body is available.

As such, I think that we need to wake the consumer and call `proxy_on_request_body()` for each chunk, passing the size of the all available data, so that it can make a decision at each point.

We could possibly add `StopAndBufferUntilEndOfStream` if needed, but that's more of an optimization that would work only in some cases, than a generic solution.

gbrail: For bodies larger than the buffer limit, it's true that the callback will never arrive because the HTTP machinery will reject the request. But this won't help if someone wants to only read the first X bytes. So let me go back to calling the callback every time and documenting the effect.

source/extensions/common/wasm/context.cc
```diff
@@ -1603,7 +1623,7 @@ Http::FilterHeadersStatus Context::decodeHeaders(Http::RequestHeaderMap& headers
 Http::FilterDataStatus Context::decodeData(Buffer::Instance& data, bool end_stream) {
   request_body_buffer_ = &data;
   end_of_stream_ = end_stream;
-  auto result = onRequestBody(data.length(), end_stream);
```
PiotrSikora: Isn't this value still valid, even in the buffered case?

gbrail: If buffering was enabled, then we pass the length of the buffered data to the callback, and this value otherwise, so it's only accurate when buffering is disabled.

In the "on*Body" method we call "getBuffer" to figure out which one to get the length of. I figured that was cleaner than putting additional conditional logic in "on*Body" to determine whether to use this length or the length of the buffer from "getBuffer".

source/extensions/common/wasm/context.cc
```diff
@@ -1209,22 +1215,28 @@ Http::FilterHeadersStatus Context::onRequestHeaders() {
...
 }
 
-Http::FilterDataStatus Context::onRequestBody(int body_buffer_length, bool end_of_stream) {
+Http::FilterDataStatus Context::onRequestBody(bool end_of_stream) {
   if (!wasm_->on_request_body_) {
     return Http::FilterDataStatus::Continue;
   }
   DeferAfterCallActions actions(this);
+  const auto buf = getBuffer(BufferType::HttpRequestBody);
+  const int bodyLen = buf == nullptr ? 0 : buf->length();
```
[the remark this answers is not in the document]
gbrail: Will fix.

source/extensions/common/wasm/context.cc
```diff
@@ -1261,6 +1273,8 @@ Http::FilterHeadersStatus Context::onResponseHeaders() {
     onCreate(root_context_id_);
     in_vm_context_created_ = true;
   }
+  // On the response body we have a new buffer
+  buffering_body_ = false;
```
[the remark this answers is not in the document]
gbrail: Will fix.

source/extensions/common/wasm/context.cc
```diff
@@ -1270,22 +1284,28 @@ Http::FilterHeadersStatus Context::onResponseHeaders() {
...
 }
 
-Http::FilterDataStatus Context::onResponseBody(int body_buffer_length, bool end_of_stream) {
+Http::FilterDataStatus Context::onResponseBody(bool end_of_stream) {
   if (!wasm_->on_response_body_) {
     return Http::FilterDataStatus::Continue;
   }
   DeferAfterCallActions actions(this);
+  const auto buf = getBuffer(BufferType::HttpResponseBody);
+  const int bodyLen = buf == nullptr ? 0 : buf->length();
```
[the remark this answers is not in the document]
gbrail: Will fix.

source/extensions/common/wasm/context.cc
PiotrSikora: Yeah, I saw that... But for some reason I assumed that the `decodeData()` call would have the buffered body buffer passed here, which turns out it's not the case. I agree that handling the extra login in `getBuffer` is cleaner.

source/extensions/common/wasm/context.cc
```diff
@@ -1209,22 +1215,28 @@ Http::FilterHeadersStatus Context::onRequestHeaders() {
...
 }
 
-Http::FilterDataStatus Context::onRequestBody(int body_buffer_length, bool end_of_stream) {
+Http::FilterDataStatus Context::onRequestBody(bool end_of_stream) {
   if (!wasm_->on_request_body_) {
     return Http::FilterDataStatus::Continue;
   }
   DeferAfterCallActions actions(this);
+  const auto buffer = getBuffer(BufferType::HttpRequestBody);
+  const auto body_len = (buffer == nullptr) ? 0 : buffer->length();
```
PiotrSikora: Style: `body_len` => `buffer_length`, otherwise we use 2 terms to describe the same buffer.

source/extensions/common/wasm/context.cc
```diff
@@ -1270,22 +1282,28 @@ Http::FilterHeadersStatus Context::onResponseHeaders() {
...
 }
 
-Http::FilterDataStatus Context::onResponseBody(int body_buffer_length, bool end_of_stream) {
+Http::FilterDataStatus Context::onResponseBody(bool end_of_stream) {
   if (!wasm_->on_response_body_) {
     return Http::FilterDataStatus::Continue;
   }
   DeferAfterCallActions actions(this);
+  const auto buffer = getBuffer(BufferType::HttpResponseBody);
+  const auto body_len = (buffer == nullptr) ? 0 : buffer->length();
```
PiotrSikora: (same here)

test/extensions/filters/http/wasm/test_data/body_cpp.cc
```diff
@@ -0,0 +1,76 @@
...
+  return FilterHeadersStatus::Continue;
+}
+
+FilterHeadersStatus ExampleContext::onResponseHeaders(uint32_t) {
+  test_op_ = getResponseHeader("x-test-operation")->toString();
+  return FilterHeadersStatus::Continue;
+}
+
+void ExampleContext::logBody(BufferType bt) {
+  size_t bufferedSize;
```
PiotrSikora: Nit: `snake_case` for local variables please.

PiotrSikora:
Actually, I'm just going to merge it now, since I have a follow-up that touches the same areas of the code, so I can cleanup style there.

gbrail:
Thank you and for directing me on the style!

PR #438, opened 2020-02-28
Add proxy_abi_version_X_Y_Z and mark what we have as v0.1.0.

PiotrSikora:


PiotrSikora:
cc @jplevyak @mandarjog @kyessenov @bianpengyuan We're going to bikeshed the ABI over the next few weeks, so some things are going to change, but let's mark what we have now as v0.1.0, so that we have some frozen ABI version that Istio 1.5.0 is going to conform to.

Any thoughts / reasons against doing that right now?

mandarjog:
Thanks for brining it up. 
It is important that the first version of Istio that we invite people to write plugins against be versioned. 

1. This is simple version check and Module load will fail if there is a mismatch. I think it is great for our first release.
2. The abi is large, Should we consider abi groups? 
     Context access, network access, stats etc?

PiotrSikora:
> 1. This is simple version check and Module load will fail if there is a mismatch. I think it is great for our first release.

We need to add support for backward-compatible changes later, but that's not needed in the first version... and the next one is going to be a breaking change, anway.

> 2. The abi is large, Should we consider abi groups?
>    Context access, network access, stats etc?

It's not that large, and while we're going to be adding new features in the future, once we do all the bikeshedding during upstreaming and WASI, the functions are going to stay stable, so I'd be against prematurely adding ABI groups.

PR #321, opened 2019-11-22
merge upstream/master to wasm

lizan:
Description:
This is currently blocking istio/istio from picking any upstream API changes because of config_dump protobuf changed.

Risk Level:
Testing:
Docs Changes:
Release Notes:

PR #440, opened 2020-03-01
Restore and fix the Wasm Service.

PiotrSikora:
This change restores Wasm Service, which was accidentally removed
in a bad merge (#321).

It also fixes a number of issuses:

1. Wasm Services were started before Cluster Manager was initialized,
   which crashed Envoy when Wasm Service tried to dispatch a callout.

2. Wasm Services (per-thread) were not stored and they were getting
   out of scope, so they were immediately destroyed.

3. Wasm Services (singleton) were started twice.

### DIFF
--- a/include/envoy/server/wasm.h
+++ b/include/envoy/server/wasm.h
@@ -5,6 +5,7 @@
 #include "envoy/common/exception.h"
 #include "envoy/common/pure.h"
 #include "envoy/event/dispatcher.h"
+#include "envoy/thread_local/thread_local.h"
 
 #include "common/common/logger.h"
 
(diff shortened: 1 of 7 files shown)


PiotrSikora:
@jplevyak this change breaks `CreateWasmFromWASMPerThread` test because Wasm Service VMs are now stored and aren't destroyed at the end of the test. Any ideas how to fix it?

PiotrSikora:
Note: This messed up the TLS slot when using per-thread Wasm Service. Any ideas on what might be the issue?

PiotrSikora:
FWIW, regular tests fail with:
```
test/mocks/thread_local/mocks.h:55: Failure
Expected: (index_) < (parent_.data_.size()), actual: 0 vs 0
Stack trace:
  0x2e0cc2e: Envoy::ThreadLocal::MockInstance::SlotImpl::~SlotImpl()
  0x2e0ccfc: Envoy::ThreadLocal::MockInstance::SlotImpl::~SlotImpl()
  0x1adbd0f: std::__1::default_delete<>::operator()()
  0x1adbccf: std::__1::unique_ptr<>::reset()
  0x1adbc49: std::__1::unique_ptr<>::~unique_ptr()
  0x1adb855: Envoy::Extensions::Wasm::WasmFactory::~WasmFactory()
  0x1adb815: Envoy::Registry::RegisterFactory<>::~RegisterFactory()
  0x7f6e74494ff8: (unknown)
```
as if somehow the SlotAllocator doesn't know about those slots.

PR #442, opened 2020-03-03
wasm: add proxy_set_buffer_bytes.

PiotrSikora:
This function can be used to:

1. Prepend data to the buffer using proxy_set_buffer_bytes(0, 0, data);

2. Append data to the buffer using proxy_set_buffer_bytes(MAX, 0, data);

3. Replace data in the buffer using proxy_set_buffer_bytes(0, MAX, data);

4. Remove data from the buffer using proxy_set_buffer_bytes(0, MAX, "");

While there, allow using MAX in proxy_get_buffer_bytes().


api/wasm/cpp/proxy_wasm_externs.h
```diff
@@ -94,6 +94,8 @@ extern "C" WasmResult proxy_get_buffer_bytes(BufferType type, uint32_t start, ui
                                              const char** ptr, size_t* size);
 extern "C" WasmResult proxy_get_buffer_status(BufferType type, size_t* length_ptr,
                                               uint32_t* flags_ptr);
+extern "C" WasmResult proxy_set_buffer(BufferType type, uint32_t start, uint32_t length,
+                                       const char* ptr, size_t size, uint32_t flags);
```
PiotrSikora: I'm not sure if this should be old-style `proxy_set_buffer_bytes` or new-style `proxy_set_buffer`? What do you think, @jplevyak?

PiotrSikora:
cc @gbrail

api/wasm/cpp/proxy_wasm_externs.h
jplevyak: Let's be consistent and make this proxy_set_buffer_bytes and then do the ABI changes in one go.

source/extensions/common/wasm/context.cc
```diff
@@ -829,6 +829,33 @@ const Buffer::Instance* Context::getBuffer(BufferType type) {
   return nullptr;
 }
 
+WasmResult Context::setBuffer(BufferType type, std::function<void(Buffer::Instance&)> callback) {
```
jplevyak: With the new changes, the Context:: API is going to be host independent.  That is, it should not reference Buffer::Instance for the top level virtual.   The envoy specific Context will override the API with a version that knows about Buffer::Instance.

source/extensions/common/wasm/exports.cc
```diff
@@ -498,6 +499,48 @@ Word get_buffer_status(void* raw_context, Word type, Word length_ptr, Word flags
...
+  auto buffer = context->getBuffer(static_cast<BufferType>(type.u64_));
+  if (!buffer) {
+    return wasmResultToWord(WasmResult::NotFound);
+  }
+  auto data = context->wasmVm()->getMemory(data_ptr.u64_, data_size.u64_);
+  if (!data) {
+    return wasmResultToWord(WasmResult::InvalidMemoryAccess);
+  }
+  WasmResult result = WasmResult::InternalFailure;
+  if (context->setBuffer(static_cast<BufferType>(type.u64_),
```
jplevyak: This should just be a thin wrapper to get the string_view(s) and do other format conversions then call a function on Context to do the real work.

test/extensions/filters/http/wasm/test_data/body_cpp.cc
```diff
@@ -34,27 +34,61 @@ FilterHeadersStatus ExampleContext::onResponseHeaders(uint32_t) {
   return FilterHeadersStatus::Continue;
 }
 
-void ExampleContext::logBody(BufferType bt) {
```
jplevyak: This isn't go, s/bt/buffer_type/

test/extensions/filters/http/wasm/test_data/body_cpp.cc
```diff
@@ -34,27 +34,61 @@ FilterHeadersStatus ExampleContext::onResponseHeaders(uint32_t) {
...
-  size_t bufferedSize;
+void ExampleContext::logBody(BufferType type) {
```
jplevyak: type->buffer_type for consistency or move them all to "type"

test/extensions/filters/http/wasm/test_data/body_cpp.cc
PiotrSikora: That's and old code that I'm fixing here.

test/extensions/filters/http/wasm/test_data/body_cpp.cc
PiotrSikora: They are all `type` here.

api/wasm/cpp/proxy_wasm_externs.h
PiotrSikora: Done.

Should I also add a no-op `proxy_set_buffer_status` in order to include it in v0.1.0 ABI? (there is no easy way to emit `end_of_stream` in HTTP body, as far as I can tell).

source/extensions/common/wasm/context.cc
PiotrSikora: Per discussion off-line, I'm going to leave this as-is, since `getBuffer` is currently also using `Buffer::Instance`, so we'll need to rewrite both for the new API anyway.

gbrail:
Excellent -- thanks!

Issue #451, opened 2020-03-06
Stats purge: stats accumulate indefinitely 


mandarjog:
envoy impl of proxy wasm uses one stats Scope. The result Is that stats accumulate indefinitely even if they are no longer being updated. 

In mixer there was a periodic purge of stats which would ensure that zombie stats do not stick around.

We either want an api or if that’s not possible then automatic behavior that purges stats if requested.

One option is to create a sub scope for a filter and then only purge that.

mandarjog:
@jplevyak @douglas-reid

kyessenov:
How serious is this leak? I imagine it's probably worst at the hub in the hub-and-spoke pattern?

mandarjog:
Yes at ingress in a large system like serverless/knative.

jplevyak:
Maybe we can rope @jmarantz in on this one.  According to the estimates we made in meetings the amount memory used should be small.  We use a single Scope and a single StatNameSet which is managed via a shared_ptr in each Wasm object.   When you define a metric, it creates the Stat returns a token to the VM.  We would have to purge those tokens from the VM and from the Wasm object.  Once that was done, in theory, the Stat should be garbage collected, but we will need to talk to @jmarantz to be sure.

We could add a releaseMetric() call which would purge the entry from the Wasm object and you could use in the plugin code.

jplevyak:
Please see the discussion on #452 @mandarjog 
We could enable purging of all stats on a VM basis by recreating the Scope object.   You would then need to defineMetric() again for all metrics you want to use.  Is this something you would be interested in?

jmarantz:
Actually the Counter& returned in this scenario would be the same. ie I believe this test will work:
```
  ScopePtr s1 = stats_store_.createScope("scope");
  ScopePtr s2 = stats_store_.createScope("scope"); // second scope has same name
  EXPECT_NE(s1.get(), s2.get()); // two different scope objects.
  StatNameManagedStore storage("foo", stats_store_.symbolTable());
  Counter& c1 = s1->counterFromStatName(storage.statName());
  Counter& c2 = s2->counterFromStatName(storage.statName());
  EXPECT_EQ(&c1, &c2); // Ultimate the stats are name-matched in the allocator, and they'll both be called "scope.foo".
```

So I think you can share the maps populated from defineMetric.

mandarjog:
@jplevyak we will have to make adjustments to code to call define metrics again, and we will need api support.
[...]

PR #452, opened 2020-03-10
Use StatNameManagedStorage for creating StatName(s).

jplevyak:

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1407,7 +1407,7 @@ void Context::onGrpcReceiveTrailingMetadata(uint32_t token, Http::HeaderMapPtr&&
 }
 
 WasmResult Context::defineMetric(MetricType type, absl::string_view name, uint32_t* metric_id_ptr) {
-  auto stat_name = wasm_->stat_name_set_->add(name);
+  auto stat_name = wasm_->stat_name_table_->intern(name);
   if (type == MetricType::Counter) {
     auto id = wasm_->nextCounterMetricId();
     auto c = &wasm_->scope_->counterFromStatName(stat_name);
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1407,7 +1407,8 @@ void Context::onGrpcReceiveTrailingMetadata(uint32_t token, Http::HeaderMapPtr&&
 }
 
 WasmResult Context::defineMetric(MetricType type, absl::string_view name, uint32_t* metric_id_ptr) {
-  auto stat_name = wasm_->stat_name_table_->intern(name);
+  Stats::StatNameManagedStorage storage(name, wasm()->scope_->symbolTable());
+  auto stat_name = storage.statName();
   if (type == MetricType::Counter) {
     auto id = wasm_->nextCounterMetricId();
     auto c = &wasm_->scope_->counterFromStatName(stat_name);
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1407,8 +1407,9 @@ void Context::onGrpcReceiveTrailingMetadata(uint32_t token, Http::HeaderMapPtr&&
 }
 
 WasmResult Context::defineMetric(MetricType type, absl::string_view name, uint32_t* metric_id_ptr) {
+  // TODO: Consider rethinking the scoping policy as it does not help in this case.
   Stats::StatNameManagedStorage storage(name, wasm()->scope_->symbolTable());
-  auto stat_name = storage.statName();
+  Stats::StatName stat_name = storage.statName();
   if (type == MetricType::Counter) {
     auto id = wasm_->nextCounterMetricId();
     auto c = &wasm_->scope_->counterFromStatName(stat_name);
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -104,19 +104,6 @@ const uint8_t* decodeVarint(const uint8_t* pos, const uint8_t* end, uint32_t* ou
 
 } // namespace
 
-// TODO: consider moving this into StatNameSet as it currently requires 3 table lookups to insert a
-// new entry.
-Stats::StatName StatNameTable::intern(absl::string_view name) {
-  Thread::LockGuard lock(mutex_);
-  auto none = Stats::StatName();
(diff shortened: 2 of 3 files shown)


jplevyak:
@mandarjog

jplevyak:
https://github.com/envoyproxy/envoy-wasm/issues/451

source/extensions/common/wasm/wasm.cc
```diff
@@ -104,6 +104,19 @@ const uint8_t* decodeVarint(const uint8_t* pos, const uint8_t* end, uint32_t* ou
 
 } // namespace
 
+// TODO: consider moving this into StatNameSet as it currently requires 3 table lookups to insert a
+// new entry.
+Stats::StatName StatNameTable::intern(absl::string_view name) {
```
jmarantz: i need to chat with you about this; I don't think this is the best pattern for this architecture, and you can tell from other uses of StatNameSet in the codebase that this is not what it's designed for.

I'm in a meeting but I"ll ping you in a few minutes

jplevyak: I think we have a short term question which is: is this the source of the memory issue that Istio is having and a longer term one which is what is the right approach for Wasm Stats.

I agree that this isn't what we want in the long term.

source/extensions/common/wasm/wasm.h
```diff
@@ -55,6 +49,20 @@ using VmConfig = envoy::extensions::wasm::v3::VmConfig;
 using WasmForeignFunction =
     std::function<WasmResult(Wasm&, absl::string_view, std::function<void*(size_t size)>)>;
 
+// Intern table for StatName(s).
+class StatNameTable {
+public:
+  explicit StatNameTable(std::unique_ptr<Stats::StatNameSet> stat_name_set)
+      : stat_name_set_(std::move(stat_name_set)) {}
+
+  Stats::StatName intern(absl::string_view str) ABSL_LOCKS_EXCLUDED(mutex_);
```
jmarantz: we can drop this class right?

source/extensions/common/wasm/context.cc
```diff
@@ -1407,7 +1407,8 @@ void Context::onGrpcReceiveTrailingMetadata(uint32_t token, Http::HeaderMapPtr&&
 }
 
 WasmResult Context::defineMetric(MetricType type, absl::string_view name, uint32_t* metric_id_ptr) {
-  auto stat_name = wasm_->stat_name_set_->add(name);
+  Stats::StatNameManagedStorage storage(name, wasm()->scope_->symbolTable());
```
jmarantz: What's the lifetime of wasm()->scope_? Should we create a scope for the VM instead so you can drop stats that are no longer used in the VM?

I'm pretty sure you can have multiple Scope objects with the same name, and that the underlying stats will be shared if they have equal StatNames.

source/extensions/common/wasm/wasm.h

source/extensions/common/wasm/context.cc
jplevyak: I wasn't sure how the storage worked, but we have 1 VM per worker which have very similar lifetimes and Stats.  I am creating a single Scope and giving a std::shared_ptr to each of the VMs so that when the last one is deleted the Scope is deleted.  We could allow individual VMs to purge all stats if we gave each one a separate Scope.   However, we currently do not have that API, so it isn't clear to me what the tradeoff would be between sharing a single Scope for all VMs with the same lifetime and code (stats) or having a separate Scope for each.    We do create a separate Scope for VMs with different lifetimes and code already.

jmarantz: When do you *want* to reap stale stats?

jplevyak: I would prefer to reap them when the last VM using them exists.  *If* we have to add support for clearing the stats periodically on running VMs I am willing to do so, but I would rather pursue other ways to limit memory  growth first.  @mandarjog is requesting the reaping because of the memory usage, but I am hoping we can get it down without that feature.

source/extensions/common/wasm/wasm.cc
jplevyak: Discussing offline.  Let's see if we can get this in as a patch then work on the long term plan.

source/extensions/common/wasm/context.cc
```diff
@@ -1407,7 +1407,8 @@ void Context::onGrpcReceiveTrailingMetadata(uint32_t token, Http::HeaderMapPtr&&
...
+  auto stat_name = storage.statName();
```
jmarantz: s/auto/Stats::StatName/

https://google.github.io/styleguide/cppguide.html#Type_deduction

the type of this is obvious to me personally but will not necessarily be to readers of this code, as it is not derived from context. The type is not onerous either so I think the guideline clearly states that explicit type is preferred here.

source/extensions/common/wasm/context.cc
jmarantz: suggest dropping a TODO now to rethink the scoping policy. I think the current structure is not benefiting you.

source/extensions/common/wasm/context.cc
jplevyak: Done: nit and added a TODO.

source/extensions/common/wasm/wasm.cc
```diff
@@ -113,8 +113,7 @@ Wasm::Wasm(absl::string_view runtime, absl::string_view vm_id, absl::string_view
       time_source_(dispatcher.timeSource()), vm_configuration_(vm_configuration),
       wasm_stats_(WasmStats{
           ALL_WASM_STATS(POOL_COUNTER_PREFIX(*scope_, absl::StrCat("wasm.", runtime, ".")),
-                         POOL_GAUGE_PREFIX(*scope_, absl::StrCat("wasm.", runtime, ".")))}),
-      stat_name_set_(scope_->symbolTable().makeSet("Wasm").release()) {
+                         POOL_GAUGE_PREFIX(*scope_, absl::StrCat("wasm.", runtime, ".")))}) {
```
jmarantz: I assume it's OK to take a bunch of locks here, as this does not happen in the request path; is that right?

You can easily code around this if you need to, and I've been mulling over creating a new family of stats-macros for instantiating stats without taking locks.

jplevyak: Typically this is done in the configure path which is coming out of the dispatcher, so same thread, but not (typically) in the request path.

PR #453, opened 2020-03-10
Fix memory leak in TCP

gargnupur:
Signed-off-by: gargnupur <gargnupur@google.com>

Description: Fix memory leak in TCP
Ref: https://github.com/istio/istio/issues/22007
Risk Level: Low
Testing: Tested e2e in istio/proxy
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -281,6 +281,16 @@ WasmResult Context::setTickPeriod(std::chrono::milliseconds tick_period) {
   return WasmResult::Ok;
 }
 
+void Context::onCloseTCP(){
+  if(tcp_connection_closed_){
+    return;
+  }
+  tcp_connection_closed_ = true;
+  onDone();
+  onLog();
+  onDelete();
+}
+
 uint64_t Context::getCurrentTimeNanoseconds() {
   return std::chrono::duration_cast<std::chrono::nanoseconds>(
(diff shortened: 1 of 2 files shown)


source/extensions/common/wasm/context.cc
```diff
@@ -281,6 +281,16 @@ WasmResult Context::setTickPeriod(std::chrono::milliseconds tick_period) {
   return WasmResult::Ok;
 }
 
+void Context::onCloseTCP() {
+  if (tcp_connection_closed_) {
+    return;
+  }
+  tcp_connection_closed_ = true;
+  onDone();
```
PiotrSikora: We should check the result of `onDone` and postpone `onLog` and `onDelete` until other extensions agree, shouldn't we, @jplevyak?

jplevyak: That logic is specifically for RootContexts and is already handled elsewhere.  The stream contexts are required to return "true" from the onDone call (the SDK ensure that)..   It is expected that any long lived operations which could outlast the stream will be handled by the RootContext.

gargnupur: So, in that case.. no need to handle result of onDone.. Guys, if you agree with the PR, please approve and merge it. Don't have merge rights for envoy-wasm...

Issue #458, opened 2020-03-18
Streaminfo not accessible in async callback


bianpengyuan:
When making async call inside `onRequestHeader` (and I think the other event handlers as well), pointers to streaminfo's data struct (e.g. request headers) will be reset after return from `onRequestHeader`, which makes those data structure not accessible in async call's callback. Seems like `SetEffectiveContext` does not have access to those data. We probably should keep those pointers if module event handlers return StopIteration.

jplevyak:
What other aspects other than the headers do you need ?

bianpengyuan:
I only need headers for the sample plugin. I think your PR already covers all other aspects that I can think of.

PR #459, opened 2020-03-18
Ensure that headers/trailers/metadata are available after StopIteration

jplevyak:
until "continue" is called.  Includes test.


bianpengyuan:
Fix https://github.com/envoyproxy/envoy-wasm/issues/458

source/extensions/common/wasm/context.cc
```diff
@@ -1660,33 +1664,59 @@ void Context::onDelete() {
   }
 }
 
+void Context::continueRequest() {
+  if (decoder_callbacks_)
```
PiotrSikora: Nit: missing brackets.

source/extensions/common/wasm/context.cc
```diff
@@ -1660,33 +1664,59 @@ void Context::onDelete() {
...
+    decoder_callbacks_->continueDecoding();
+  request_headers_ = nullptr;
+  request_body_buffer_ = nullptr;
+  request_trailers_ = nullptr;
+  request_metadata_ = nullptr;
+}
+
+void Context::continueResponse() {
+  if (encoder_callbacks_)
```
PiotrSikora: (same here)

source/extensions/common/wasm/context.cc
jplevyak: Done.

source/extensions/common/wasm/context.cc
jplevyak: Done

PR #461, opened 2020-03-20
remove on create initial metadata

bianpengyuan:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


bianpengyuan:
We don't need to actually. There is already a option in GrpcService that adds the initial metadata. https://github.com/envoyproxy/envoy/blob/9186ebf374b1bd7d7c65ed1cec748377f88ec0a7/api/envoy/api/v2/core/grpc_service.proto#L225

PiotrSikora:
While the initial metadata can indeed be set in the GrpcService proto, you need to modify proto for each request and you're mixing service and per-stream metadata in a weird way, IMHO.

I think we should still add the initial metadata to the calls, like @jplevyak mentioned, and as we discussed in the past. What do you think?

bianpengyuan:
Yeah that makes sense to me. John talked to me offline and he is going to work on the PR to add that param. He mentioned the other reason though that GrpcService proto is Envoy specific and we want to make ABI proxy neutral.

PR #466, opened 2020-03-26
Move to use proxy-wasm-cpp-sdk for the envoy code.

jplevyak:
Followup PRs will move the .wasm code generation over
and will  move to use proxy-wasm-cpp-host.

### DIFF
--- a/bazel/external/proxy_wasm_cpp_host.BUILD
+++ b/bazel/external/proxy_wasm_cpp_host.BUILD
@@ -1,0 +1,54 @@
+licenses(["notice"])  # Apache 2
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "include",
+    hdrs = [
+        "include/proxy-wasm/compat.h",
+        "include/proxy-wasm/context.h",
+        "include/proxy-wasm/exports.h",
+        "include/proxy-wasm/null.h",
+        "include/proxy-wasm/null_plugin.h",
+        "include/proxy-wasm/null_vm.h",
+        "include/proxy-wasm/null_vm_plugin.h",
+        "include/proxy-wasm/v8.h",
+        "include/proxy-wasm/wasm.h",
+        "include/proxy-wasm/wasm_api_impl.h",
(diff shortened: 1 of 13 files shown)


jplevyak:
@yxue

bazel/external/proxy_wasm_cpp_host.BUILD
```diff
@@ -0,0 +1,54 @@
+licenses(["notice"])  # Apache 2
```
PiotrSikora: I'm confused as to why do we need this? Doesn't [proxy-wasm-cpp-host](https://github.com/proxy-wasm/proxy-wasm-cpp-host) provide native BUILD file already?

jplevyak: The problem is that the dependencies for sub-repos are injected from the main repos WORKSPACE in bazel.  This means that if there are changes in Envoy repositories in WORKSPACE, then the SDK BUILD file would have to change.  By storing the BUILD file for proxy-wasm-cpp-sdk in Envoy proper we can break that dependency.  This also saves the SDK BUILD file from being dependent on the Envoy WORKSPACE so that it can change independently in ways that may not match with the Enovy WORKSPACE.

PiotrSikora: I disagree. I don't see why changes in Envoy should influence `WORKSPACE` or `BUILD` files in `proxy-wasm-cpp-{host,sdk}`. That's definitely not the case for any of the other 30+ Envoy dependencies. Also, in Bazel-world, the main project needs to add all transitive dependencies to its `WORKSPACE`, so if anything, it would be the other way around.

Having said that, I see that right now the Wasm runtimes are excluded from the `BUILD` file in [proxy-wasm-cpp-host](https://github.com/proxy-wasm/proxy-wasm-cpp-host), so I guess we can do this for the time being... But I don't think this `BUILD` file should stay long-term.

lizan: +1

PR #476, opened 2020-04-06
wasm abi: tls certificate information retrieval

Shikugawa:
Signed-off-by: shikugawa <Shikugawa@gmail.com>

Description: This is only a proposal implementation. I added instructions to retrieve local/peer certificate information. It is required to retrieve on AuthN filter wasm implementation. If there is no implementation to retrieve cert info via wasm VM, we should link upstream envoy connection info. It may cause the complexity of dependencies of AuthN wasm implementation.

https://github.com/istio/istio/issues/15772

Risk Level: Low
Testing: N/A
Docs Changes: Required
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -33,6 +33,7 @@
 #include "absl/container/flat_hash_map.h"
 #include "absl/container/node_hash_map.h"
 #include "absl/strings/str_cat.h"
+#include "absl/strings/str_join.h"
 #include "absl/synchronization/mutex.h"
 #include "eval/eval/field_access.h"
 #include "eval/eval/field_backed_list_impl.h"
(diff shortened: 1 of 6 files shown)


source/extensions/common/wasm/context.cc
```diff
@@ -1087,6 +1087,44 @@ const Network::Connection* Context::getConnection() const {
   return nullptr;
 }
 
+std::string spanToString(const absl::Span<const std::string>& span) {
```
dio, line 1090: Can we do [`absl::StrJoin(span, ",")`](https://github.com/abseil/abseil-cpp/blob/d43b7997c0ca0f3312a51d1057fb73cfe934703b/absl/strings/str_join.h#L172-L173)?

PiotrSikora:
@Shikugawa thanks for the PR, but could you send a PR with the ABI changes to https://github.com/proxy-wasm/spec, or just create an issue there to discuss this feature?

Ultimately, I don't think that we need to have a separate function for each buffer/map, and this could be exposed either as a raw certificate via `proxy_get_buffer(TlsCertificate{Local,Peer}, ...)` or as a map with extracted values via `proxy_get_map(TlsCertificate{Local,Peer}, ...)`. 

cc @jplevyak

kyessenov:
Why do we need this if we can use `getValue({"connection", "subject_local_certificate"})` from https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/rbac_filter#condition?

Shikugawa:
@kyessenov If we use this, we must build `rbac_filter` with emscripten. But we'd not like to do what the number of dependencies related to upstream envoy will grow up. So we need this functionality.

kyessenov:
@Shikugawa I don't follow. getValue lets you add more getters without bloating the API and we already added a bunch for TLS info for telemetry. There's no extra dependencies on rbac filter, it just uses the underlying code to match the property names.

Shikugawa:
@kyessenov Thanks. I'll try it with this. But, this functionality may be required in the future. So I think that continuing a discussion with this is useful.

kyessenov:
Sure, please take a look and we can decide how to grow the set attributes. I just don't want two different ways to get the same things.

Shikugawa:
@kyessenov I tried it with `getValue`. But it has less functionality to solve our problem such as we can't retrieve whether peer certificate is presented. So still we need this.

kyessenov:
@Shikugawa See `connection.mtls`: Indicates whether TLS is applied to the downstream connection and the peer ceritificate is presented. This is not entirely possible on the client by the way, there is my long failed attempt to implement it in envoy repo.

source/extensions/common/wasm/exports.h
```diff
@@ -80,9 +80,10 @@ Word set_effective_context(void* raw_context, Word context_id);
 Word done(void* raw_context);
 Word call_foreign_function(void* raw_context, Word function_name, Word function_name_size,
                            Word arguments, Word warguments_size, Word results, Word results_size);
+Word get_peer_certificate_info(void* raw_context, Word value_ptr_ptr, Word value_size_ptr);
```
jplevyak, line 83: Instead of adding this as a core ABI, let's add it to getProperties.

PR #477, opened 2020-04-07
Use proxy-wasm-cpp-host in envoy-wasm.

jplevyak:


api/wasm/cpp/proxy_wasm_common.h
```diff
@@ -65,14 +65,16 @@ enum class HeaderMapType : int32_t {
   HttpCallResponseTrailers = 7,    // Immutable
   MAX = 7,
 };
-enum class BufferType : int32_t {
+enum class WasmBufferType : int32_t {
```
yxue, line 68: Why not use the header files in proxy-wasm/proxy-wasm-cpp-sdk?

jplevyak: I need to move all the tests to use that repo and I just haven't got around to it.

That will be my next task unless you want to do it ;)

test/extensions/common/wasm/test_data/test_cpp_null_plugin.cc
```diff
@@ -1,38 +1,15 @@
...
+NullPluginRegistry* context_registry_;
 } // namespace CommonWasmTestCpp
 
-/**
- * Config registration for a Wasm filter plugin. @see NamedHttpFilterConfigFactory.
- */
-class CommonWasmTestPluginFactory : public NullVmPluginFactory {
-public:
-  CommonWasmTestPluginFactory() {}
+RegisterNullVmPluginFactory register_common_wasm_test_cpp_plugin("CommonWasmTestCpp", []() {
```
yxue, line 10: Just curious, how we register the factory now? I saw you comment the following code. 

```
static Registry::RegisterFactory<CommonWasmTestPluginFactory, NullVmPluginFactory> register_;
```

test/extensions/common/wasm/test_data/test_cpp_null_plugin.cc
```diff
@@ -1,38 +1,15 @@
...
-namespace Extensions {
-namespace Common {
-namespace Wasm {
-namespace Null {
-namespace Plugin {
+namespace proxy_wasm {
+namespace null_plugin {
 namespace CommonWasmTestCpp {
-ThreadSafeSingleton<NullPluginRegistry> null_plugin_registry_;
+NullPluginRegistry* context_registry_;
```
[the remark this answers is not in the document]
jplevyak: This is a POD so it is initialized from DSS (in this case zero-initialized).  This occurs before any initializers run so it is safe.

test/extensions/common/wasm/test_data/test_cpp_null_plugin.cc
jplevyak: The new registration is by global variable constructor, just like the old one but less complicated because it does not need to be as general as the Envoy version.  In this case, the constructor for RegisterNullVmPluginFactory inserts the provided function into a table which is hung off a pointer (to avoid the initialization sequencing issue).   Since these are global initializers which are single threaded and since table insertions can occur in any order it is safe.

PR #478, opened 2020-04-09
Prevent reentrant HttpCall callbacks.

jplevyak:


source/extensions/common/wasm/context.cc
```diff
@@ -1589,7 +1589,14 @@ void Context::setEncoderFilterCallbacks(Envoy::Http::StreamEncoderFilterCallback
...
 
-void Context::onHttpCallSuccess(uint32_t token, Envoy::Http::ResponseMessagePtr& response) {
+void Context::onHttpCallSuccess(uint32_t token, Envoy::Http::ResponseMessagePtr&& response) {
+  if (proxy_wasm::current_context_ != nullptr) {
+    // We are in a reentrant call, so defer.
+    auto response_ptr = response.release();
+    wasm()->addAfterVmCallAction([this, token, response_ptr] {
+      onHttpCallSuccess(token, std::unique_ptr<Envoy::Http::ResponseMessage>(response_ptr));
+    });
+  }
```
PiotrSikora, line 1599: Shouldn't there be a `return;` at the end of each `if` statement? Otherwise, you're going to execute both code patches.

source/extensions/common/wasm/context.cc
```diff
@@ -1589,7 +1589,14 @@ void Context::setEncoderFilterCallbacks(Envoy::Http::StreamEncoderFilterCallback
   encoder_callbacks_ = &callbacks;
 }
 
-void Context::onHttpCallSuccess(uint32_t token, Envoy::Http::ResponseMessagePtr& response) {
+void Context::onHttpCallSuccess(uint32_t token, Envoy::Http::ResponseMessagePtr&& response) {
+  if (proxy_wasm::current_context_ != nullptr) {
+    // We are in a reentrant call, so defer.
+    auto response_ptr = response.release();
+    wasm()->addAfterVmCallAction([this, token, response_ptr] {
```
PiotrSikora, line 1596: Nit: wouldn't this be more readable?

```
wasm()->addAfterVmCallAction([this, token, response = response.release()] {
```

to make it clearer that we're moving this value to lambda?

source/extensions/common/wasm/context.cc
```diff
@@ -1622,6 +1634,13 @@ void Context::onGrpcReceiveWrapper(uint32_t token, ::Envoy::Buffer::InstancePtr
 
 void Context::onGrpcCloseWrapper(uint32_t token, const Grpc::Status::GrpcStatus& status,
                                  const absl::string_view message) {
+  if (proxy_wasm::current_context_ != nullptr) {
+    // We are in a reentrant call, so defer.
+    auto message_string = std::string(message);
+    wasm()->addAfterVmCallAction([this, token, status, message_string] {
```
PiotrSikora, line 1640: (same here)

```
wasm()->addAfterVmCallAction([this, token, status, message = std::string(message)] {
```

to make it clearer that we're moving this value to lambda?

source/extensions/common/wasm/context.cc
```diff
@@ -1589,7 +1589,14 @@ void Context::setEncoderFilterCallbacks(Envoy::Http::StreamEncoderFilterCallback
   encoder_callbacks_ = &callbacks;
 }
 
-void Context::onHttpCallSuccess(uint32_t token, Envoy::Http::ResponseMessagePtr& response) {
+void Context::onHttpCallSuccess(uint32_t token, Envoy::Http::ResponseMessagePtr&& response) {
+  if (proxy_wasm::current_context_ != nullptr) {
```
PiotrSikora, line 1593: Could you turn this into a function? Accessing fields across library boundaries isn't the best practice, IMHO.

jplevyak: I'll add a TODO for that.   It requires changing the -host code which is a two step dance.  I have a bunch of queued things to do there and I'll group this in.

source/extensions/common/wasm/context.cc
jplevyak: Done.

source/extensions/common/wasm/context.cc
jplevyak: google3 went from 11->17 directly so I wasn't sure if that was a 14 thing or not.  Nice to know that it is.

source/extensions/common/wasm/context.cc
jplevyak: Woops.  Done.

PR #480, opened 2020-04-10
Cleanup tidy warnings and get more tests running.

jplevyak:

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -743,8 +743,9 @@ def _proxy_wasm_cpp_sdk():
     _repository_impl(name = "proxy_wasm_cpp_sdk")
 
 def _proxy_wasm_cpp_host():
-    _repository_impl(
+    native.new_local_repository(
         name = "proxy_wasm_cpp_host",
+        path = "/home/jplevyak/work/proxy-wasm-cpp-host",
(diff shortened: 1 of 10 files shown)


bazel/repositories.bzl
```diff
@@ -743,8 +743,9 @@ def _proxy_wasm_cpp_sdk():
     _repository_impl(name = "proxy_wasm_cpp_sdk")
 
 def _proxy_wasm_cpp_host():
-    _repository_impl(
+    native.new_local_repository(
         name = "proxy_wasm_cpp_host",
+        path = "/home/jplevyak/work/proxy-wasm-cpp-host",
```
kyessenov, line 748: leftover?

jplevyak: reverted.

source/extensions/common/wasm/wasm.cc
```diff
@@ -1,6 +1,6 @@
 #include "extensions/common/wasm/wasm.h"
 
-#include <stdio.h>
+#include <cstdio>
```
kyessenov, line 3: Why do we need stdio here, do we print stdout directly?

jplevyak: These are left over from debugging.  YCM doesn't seem to report unnecessary includes, but I checked and I can remove them.

PR #486, opened 2020-04-20
fixes to filter state getters/setters

kyessenov:

- add nullptr check for upstream filter state
- retrieve flatbuffer as bytes if no field is traversed
- propagate wasm life span to setData
- make data always mutable to avoid runtime errors


kyessenov:
Need more patches. Hold on yet.

PR #489, opened 2020-04-21
Add Wasm code cache and option to fail if not cached.

jplevyak:


jplevyak:
@mandarjog @bianpengyuan

jplevyak:
Anyone know anything about the protocol_integration_tests?  I can't imagine that this PR has any effect on that test.

jplevyak:
Could someone trigger re-test?

jplevyak:
Any clue on the protocol_integration_test ? This passes locally for me and this PR doesn't change anything associated with it.

jplevyak:
This passed here: https://app.circleci.com/pipelines/github/envoyproxy/envoy-wasm/1176/workflows/7e8bb0e2-bdeb-4dd8-b023-43782ad0059f

source/extensions/common/wasm/wasm.cc
```diff
@@ -165,45 +160,87 @@ static void createWasmInternal(const VmConfig& vm_config, const PluginSharedPtr&
...
+        std::make_shared<std::unique_ptr<Config::DataSource::RemoteAsyncDataProvider>>();
+    auto fetch_callback = [vm_config, complete_cb, source, &dispatcher,
+                           remote_data_provider_holder](const std::string& code) {
+      {
+        std::lock_guard<std::mutex> guard(code_cache_mutex);
+        auto& e = (*code_cache)[vm_config.code().remote().sha256()];
+        e.in_progress = false;
+        e.code = code;
+      }
+      if (!fail_if_code_not_cached) {
```
mandarjog, line 223: Is this so that we can still enter warming state?

source/extensions/common/wasm/wasm.cc
```diff
@@ -165,45 +160,87 @@ static void createWasmInternal(const VmConfig& vm_config, const PluginSharedPtr&
...
-          return std::static_pointer_cast<WasmHandleBase>(std::make_shared<WasmHandle>(wasm));
-        },
-        vm_config.allow_precompiled(), std::move(context))));
-  };
+  auto complete_cb =
+      [cb, vm_config, plugin, scope, &cluster_manager, &dispatcher, &lifecycle_notifier,
+       root_context_for_testing_ptr = root_context_for_testing.release()](std::string code) {
+        auto root_context = std::unique_ptr<Context>(root_context_for_testing_ptr);
+        auto vm_key = proxy_wasm::makeVmKey(vm_config.vm_id(), vm_config.configuration(), code);
+        cb(std::static_pointer_cast<WasmHandle>(proxy_wasm::createWasm(
```
mandarjog, line 196: can we add some metrics here?
How many fetches, retries? Unless it is already tracked in cluster metrics.
Perhaps it would still be better to track those here.

source/extensions/common/wasm/wasm.cc
jplevyak: I have preserved the old behavior as an option.  It isn't clear to me how other people outside of istio might want to use this.  We can remove it later if there is no interest and if xDS doesn't provide a way to send back async status.

source/extensions/common/wasm/wasm.cc
jplevyak: The retries are hidden by the RemoteAsyncDataProvider.  I can tell you how many fetches there are, but as you say, that is covered by the cluster stats.  However, if you want to know how many fetches there were and the disposition thereof I can add that.   I would imagine though that given how infrequently we update the .wasm objects those stats are not going to be very interesting.   It will only be interesting if the cluster is down.

jplevyak: Think about it and tell me what you want.  I can add it here or in a follow up PR.

test/extensions/filters/http/wasm/config_test.cc
```diff
@@ -207,6 +211,201 @@ TEST_P(WasmFilterConfigTest, YamlLoadFromRemoteWASM) {
   cb(filter_callback);
 }
 
+TEST_P(WasmFilterConfigTest, YamlLoadFromRemoteWASMFailOnUncachedThenSucceed) {
+  Envoy::Extensions::Common::Wasm::clearCodeCacheForTesting(true);
+  const std::string code = TestEnvironment::readFileToStringForTest(TestEnvironment::substitute(
```
mandarjog, line 216: apparently you can use `TestEnvironment::runfilesPath(relative_path)`

source/extensions/common/wasm/wasm.cc
```diff
@@ -38,9 +15,22 @@ namespace Wasm {
...
+// NB: xDS currently does not support failing asynchronously, so we fail immediately
+// if remote Wasm code is not cached and do a background fill.
+const bool DEFAULT_FAIL_IF_NOT_CACHED = true;
+bool fail_if_code_not_cached = DEFAULT_FAIL_IF_NOT_CACHED;
+const int CODE_CACHE_SECONDS_NEGATIVE_CACHING = 10;
 
-std::atomic<int64_t> active_wasm_;
+std::atomic<int64_t> active_wasms;
+std::mutex code_cache_mutex;
+std::unordered_map<std::string, CodeCacheEntry>* code_cache = nullptr;
```
mandarjog, line 33: This is an unbounded cache. is that ok? What is the lifetime?  server or listener,filter?

test/extensions/filters/http/wasm/config_test.cc
```diff
@@ -207,6 +211,176 @@ TEST_P(WasmFilterConfigTest, YamlLoadFromRemoteWASM) {
...
+  EXPECT_CALL(init_watcher_, ready());
+  context_.initManager().initialize(init_watcher_);
+  EXPECT_EQ(context_.initManager().state(), Init::Manager::State::Initialized);
+
+  Init::ManagerImpl init_manager2{"init_manager2"};
+  Init::ExpectableWatcherImpl init_watcher2;
+
+  EXPECT_CALL(context_, initManager()).WillRepeatedly(ReturnRef(init_manager2));
+
+  Http::FilterFactoryCb cb = factory.createFilterFactoryFromProto(proto_config, "stats", context_);
```
mandarjog, line 265: Q: how is time being managed here? 
I see 2 calls to `createFilterFactoryFromProto` with different expectations, but how do we guarantee that 1st fails and 2nd succeeds? The asyncclient can return after 2nd call, no?

source/extensions/common/wasm/wasm.cc
jplevyak: The lifetime is unbounded.  SInce it is SHA based I felt like that was safe.

test/extensions/filters/http/wasm/config_test.cc
jplevyak: The way that gtest works is that expectations are dispatched based on the closest one above (before in the code path) which is still active.  In this case I add the ones I am using above the case I am testing.  I am also using Repeatedly so that it is clear which one is going to get selected.  I can add some comments, but it will require rerunning the tests which can take a very long time.

test/extensions/filters/http/wasm/config_test.cc
jplevyak: OK.  I can fix that offline.  I"ll make a note.

PR #490, opened 2020-04-22
Changed DefaultTimeout interval to 10s

dmitri-d:
It should be less than DEFAULT_ROUTE_TIMEOUT_MS of 15s to stop intermittent failures in  test/integration/protocol_integration_test.

### DIFF
--- a/test/test_common/utility.h
+++ b/test/test_common/utility.h
@@ -468,7 +468,7 @@ public:
   static std::string convertTime(const std::string& input, const std::string& input_format,
                                  const std::string& output_format);
 
-  static constexpr std::chrono::milliseconds DefaultTimeout = std::chrono::milliseconds(30000);
+  static constexpr std::chrono::milliseconds DefaultTimeout = std::chrono::milliseconds(10000);
 
   /**
    * Return a prefix string matcher.


dmitri-d:
> we're not upstreaming it

What are the plans for this repository? I see that the master branch of istio-proxy is built using this version of envoy.

PiotrSikora:
> What are the plans for this repository? I see that the master branch of istio-proxy is built using this version of envoy.

Sorry, I should have been more explicit.

The change to increase test timeout, that you're reverting in this PR, is only needed when using WAVM runtime without precompiled modules, since compilation at load-time takes forever. We're upstreaming NullVM and V8 runtimes, but not WAVM runtime, hence my comment.

@jplevyak please merge this, unless you have any objections (e.g. you don't want to break WAVM).

dmitri-d:
> Sorry, I should have been more explicit.
>
> The change to increase test timeout, that you're reverting in this PR, is only needed when using WAVM runtime without precompiled modules, since compilation at load-time takes forever. We're upstreaming NullVM and V8 runtimes, but not WAVM runtime, hence my comment.

Sorry, I meant to ask plans for the repository as a whole? Is it going to be used in the (eventual) release? Or just the runtimes you mention are going to be ported upstream and then that version of Envoy used?

PR #491, opened 2020-04-22
Added 'default' cases in switches

dmitri-d:
gcc isn't happy without those

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1161,6 +1161,7 @@ WasmResult Context::log(uint64_t level, absl::string_view message) {
     ENVOY_LOG(critical, "wasm log{}: {}", log_prefix(), message);
     return WasmResult::Ok;
   case spdlog::level::off:
+  default:
     NOT_IMPLEMENTED_GCOVR_EXCL_LINE;
   }
 }
@@ -1321,6 +1322,8 @@ Network::FilterStatus convertNetworkFilterStatus(proxy_wasm::FilterStatus status
     return Network::FilterStatus::Continue;
   case proxy_wasm::FilterStatus::StopIteration:
     return Network::FilterStatus::StopIteration;
+  default:
+    NOT_IMPLEMENTED_GCOVR_EXCL_LINE;
   }
 };
 
@@ -1330,6 +1333,8 @@ Http::FilterHeadersStatus convertFilterHeadersStatus(proxy_wasm::FilterHeadersSt
     return Http::FilterHeadersStatus::Continue;
   case proxy_wasm::FilterHeadersStatus::StopIteration:
     return Http::FilterHeadersStatus::StopIteration;
+  default:
+    NOT_IMPLEMENTED_GCOVR_EXCL_LINE;
   }
 };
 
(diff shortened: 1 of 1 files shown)


source/extensions/common/wasm/context.cc
```diff
@@ -1161,6 +1161,7 @@ WasmResult Context::log(uint64_t level, absl::string_view message) {
     ENVOY_LOG(critical, "wasm log{}: {}", log_prefix(), message);
     return WasmResult::Ok;
   case spdlog::level::off:
+  default:
```
PiotrSikora, line 1164: Not the core of your change, but `spdlog::level::off` should probably be a no-op and simply `return WasmResult::Ok`.

cc @jplevyak

jplevyak: I agree.

source/extensions/common/wasm/context.cc
```diff
@@ -1321,6 +1322,8 @@ Network::FilterStatus convertNetworkFilterStatus(proxy_wasm::FilterStatus status
     return Network::FilterStatus::Continue;
   case proxy_wasm::FilterStatus::StopIteration:
     return Network::FilterStatus::StopIteration;
+  default:
```
jplevyak, line 1325: I am not sure why we want to do this.  Not having default causes a compilation error if the enum changes since it requires the case to be exhaustive.

dmitri-d: Not having a default case results in compiler error under gcc. It's either that, or each of those methods needs to have an explicit return at the end.

jplevyak: Ah.  Haven't used gcc in a while and a vaguely remember it can change with optimization level as well.  Let me see if I can reproduce it.   What version of gcc are you using?

jplevyak: Please take a look at https://github.com/jplevyak/envoy-wasm/tree/wasm-case
where I added the NOT_IMPLEMENTED_GCOVR_EXCL_LINE outside the case and
fixed the handling of  spdlog::level::off:.

I think this addresses all the issues.  WDYT?

PiotrSikora: To be honest, I think that `NOT_IMPLEMENTED_GCOVR_EXCL_LINE` inside `switch` as a `default` case is technically a bit more correct than outside of it, even though it results in effectively the same code.

jplevyak: The problem with adding a default statement is that it prevents the compiler from issuing a warning when the cases are not exhaustive such as when new elements are added to the enum.  See https://abseil.io/tips/147   The suggested solution is the one that I am providing: to not have a default but to handle those cases outside the switch.  In this case, the range of the enum is validated in the exports.cc file for all cases, however, that isn't something which is manifest in the C++ type system, so we need to handle non-enum values somehere.

PiotrSikora: Fair point.

jplevyak:
Submitted as https://github.com/envoyproxy/envoy-wasm/pull/499

Thanx!

PR #492, opened 2020-04-22
Move wasm api v3 into extensions, remove unused v2 (not upstreamed).

jplevyak:


PiotrSikora:
Don't we need v2 for Istio, still?

jplevyak:
Do we need v2 for the next version of istio ?  This is in master.

jplevyak:
Talked to the team.  We can remove v2 for the next release, so I say we nuke v2 this for upstreaming.  Upstream had v2alpha, but Kuat already removed it, so v2 would never get used and we use typed_struct, so it is entirely within our control.

PR #494, opened 2020-04-24
build: call bazel in bazel to build test_cpp.wasm

lizan:


PiotrSikora:
cc @jplevyak

lizan:
there are much better way to do this, will raise another PR.

PR #495, opened 2020-04-27
Update proxy-wasm-cpp-host dependency for context.h comments.

jplevyak:
NOTE:  I need to update the repository location following https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/12

### DIFF
--- a/source/extensions/common/wasm/context.h
+++ b/source/extensions/common/wasm/context.h
@@ -28,10 +28,14 @@ namespace Wasm {
 #include "proxy_wasm_common.h"
 
 using proxy_wasm::BufferInterface;
+using proxy_wasm::CloseType;
 using proxy_wasm::MetricType;
 using proxy_wasm::Pairs;
 using proxy_wasm::PairsWithStringValues;
 using proxy_wasm::PluginBase;
+using proxy_wasm::SharedQueueDequeueToken;
+using proxy_wasm::SharedQueueEnqueueToken;
+using proxy_wasm::StreamType;
 using proxy_wasm::WasmBase;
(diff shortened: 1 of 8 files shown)


source/extensions/common/wasm/context.h
```diff
@@ -28,10 +28,14 @@ namespace Wasm {
 #include "proxy_wasm_common.h"
 
 using proxy_wasm::BufferInterface;
+using proxy_wasm::CloseType;
 using proxy_wasm::MetricType;
 using proxy_wasm::Pairs;
 using proxy_wasm::PairsWithStringValues;
 using proxy_wasm::PluginBase;
+using proxy_wasm::SharedQueueDequeueToken;
+using proxy_wasm::SharedQueueEnqueueToken;
```
PiotrSikora, line 37: Nit: Shouldn't this be a single type (e.g. `SharedQueueId`)? I don't think we need one type for `register` and another type for `lookup`. Also, I don't believe that we enforce who can enqueue and who can dequeue objects, and both directions (i.e. main thread pushes objects to workers, and workers push objects to main thread) are valid.

jplevyak: The different tokens were a specific request from Matt.  You are correct in that we are not enforcing who can push to whom.  In theory, a set of workers could have entries pushed to them by a main thread Wasm VM.  Do we want to enforce something like that?  I can think of valid uses cases for the main->workers example.

PiotrSikora: My point was that we don't and shouldn't enforce them, since both directions are valid.

PR #498, opened 2020-04-29
build: use transition to build wasm binary

lizan:
Use bazel transition to create wasm configuration and depends that in wasm_test.

Requires:
https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/pull/16
https://github.com/envoyproxy/envoy/pull/11022


jplevyak:
Looking good.  Very much interested in getting this in so I can convert the result of the tests.

lizan:
@jplevyak once you reviewed and merged proxy-wasm/proxy-wasm-cpp-sdk#16 this is ready to go.

jplevyak:
@lizan done https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/pull/16 merged

bazel/wasm/wasm.bzl
```diff
@@ -0,0 +1,54 @@
+def _wasm_transition_impl(settings, attr):
+    return {
+        "//command_line_option:cpu": "wasm",
+        "//command_line_option:crosstool_top": "@proxy_wasm_cpp_sdk//toolchain:emscripten",
+        # Overriding copt/cxxopt/linkopt to prevent sanitizers/coverage options leak into WASM build
```
jplevyak, line 5: Do you want to move this down 2 lines?

bazel/wasm/wasm.bzl
```diff
@@ -0,0 +1,54 @@
...
+        "binary": attr.label(mandatory = True, cfg = wasm_transition),
+        "_whitelist_function_transition": attr.label(default = "@bazel_tools//tools/whitelists/function_transition_whitelist"),
+    },
+)
+
+def wasm_cc_binary(name, **kwargs):
+    wasm_name = "_wasm_" + name
+    native.cc_binary(
+        name = wasm_name,
+        # Adding manual tag so it won't be built in non-wasm config automatically.
```
jplevyak, line 47: It would be great if we could get the .wasm for non-wasm targets which depend on them.

lizan: Yes, that's the `wasm_binary` below does, added more comment. This is the private cc_binary for the one below to depend on, so it has `_wasm_` prefix.

PR #502, opened 2020-05-02
Fetch .wasm from remote URI without depending on Listener.

jplevyak:

### DIFF
--- a/source/extensions/common/wasm/wasm.cc
+++ b/source/extensions/common/wasm/wasm.cc
@@ -23,6 +23,22 @@ struct CodeCacheEntry {
   MonotonicTime fetch_time;
 };
 
+class RemoteDataFetcherAdapter : public Config::DataFetcher::RemoteDataFetcherCallback,
+                                 public Event::DeferredDeletable {
+public:
+  RemoteDataFetcherAdapter(std::function<void(std::string cb)> cb) : cb_(cb) {}
+  ~RemoteDataFetcherAdapter() = default;
+  void onSuccess(const std::string& data) override { cb_(data); }
+  virtual void onFailure(Config::DataFetcher::FailureReason) override { cb_(""); }
+  void setFetcher(std::unique_ptr<Config::DataFetcher::RemoteDataFetcher>&& fetcher) {
+    fetcher_ = std::move(fetcher);
+  }
+
+private:
+  std::function<void(std::string)> cb_;
+  std::unique_ptr<Config::DataFetcher::RemoteDataFetcher> fetcher_;
+};
+
 const std::string INLINE_STRING = "<inline>";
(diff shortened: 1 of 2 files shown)


source/extensions/common/wasm/wasm.cc
```diff
@@ -263,16 +276,24 @@ static void createWasmInternal(const VmConfig& vm_config, const PluginSharedPtr&
...
     if (fail_if_code_not_cached) {
-      *remote_data_provider_holder = std::move(remote_data_provider);
+      auto adapter = std::make_unique<RemoteDataFetcherAdapter>(fetch_callback);
+      auto fetcher = std::make_unique<Config::DataFetcher::RemoteDataFetcher>(
+          cluster_manager, vm_config.code().remote().http_uri(), vm_config.code().remote().sha256(),
+          *adapter);
+      auto fetcher_ptr = fetcher.get();
+      adapter->setFetcher(std::move(fetcher));
+      *holder = std::move(adapter);
+      fetcher_ptr->fetch();
```
PiotrSikora, line 291: Using pointer that's moved (twice!) seems like an anti-pattern. Is there any way to make this a bit cleaner?

jplevyak: The problem here is that we need to make the  fetcher self-deleting.  We also need to use an object for the callback because that is the API for the fetcher.   Sooooo, that means creating a callback object creating the fetcher.  Storing the fetcher into the callback object, then passing the callback object into the callback function as a shared pointer.   I could switch to bare pointers which would simplify things a bit.  WDYT?  The core problem is that the API for this fetcher is not very flexible.  It is really meant to be used in the single existing use-case.

jplevyak: I could also make the function into a member function.  That would be less clear since the callback code is logically part of the createWasmInternal function, but it would eliminate some unique_ptr movement.

PiotrSikora: Aren't we the only existing use-case (i.e. couldn't we change the API for the fetcher)?

jplevyak: We could.  That would require upstreaming to envoy.  I am absolutely willing to do that, but this is being backported to 1.6 so I didn't want to make changes into core envoy as part of the backport.  We can discuss this offline.  I agree that in the long run we should change the fetcher API.  The question is what to do in the short term.

PiotrSikora: Fair enough, let's merge this as-is. Could you open tracking issue for that?

jplevyak: https://github.com/envoyproxy/envoy-wasm/issues/506

PR #511, opened 2020-05-07
Log error if a Cel error value is returned

mandarjog:
Previously we were not checking for errors.
If a top level key is missing, now a clear error is logged.

```no key named "istio" found in the activation```

### DIFF
--- a/source/extensions/common/wasm/foreign.cc
+++ b/source/extensions/common/wasm/foreign.cc
@@ -170,8 +170,13 @@ public:
         ENVOY_LOG(debug, "expr_evaluate error: {}", eval_status.status().message());
         return WasmResult::InternalFailure;
       }
+      auto value = eval_status.value();
+      if (value.IsError()) {
+        ENVOY_LOG(debug, "expr_evaluate value error: {}", value.ErrorOrDie()->message());
+        return WasmResult::InternalFailure;
+      }
       std::string result;
-      auto serialize_status = serializeValue(eval_status.value(), &result);
+      auto serialize_status = serializeValue(value, &result);
       if (serialize_status != WasmResult::Ok) {
         return serialize_status;
       }

### CODE REVIEW

File: source/extensions/common/wasm/foreign.cc [hunk 1]
```diff
@@ -170,8 +170,13 @@ class EvaluateExpressionFactory : public ExpressionFactory {
         ENVOY_LOG(debug, "expr_evaluate error: {}", eval_status.status().message());
         return WasmResult::InternalFailure;
       }
+      auto value = eval_status.value();
+      if (value.IsError()) {
```
mandarjog, line 174: @TristonianJones is this expected? I clearly saw it in practice.
`eval_status.ok() == true`  but `value.IsError() == true` ?

jplevyak: Gads

source/extensions/common/wasm/foreign.cc
```diff
@@ -170,8 +170,13 @@ class EvaluateExpressionFactory : public ExpressionFactory {
...
+        ENVOY_LOG(debug, "expr_evaluate error: {}", value.ErrorOrDie()->message());
```
jplevyak, line 175: `expr_evaluate value error" so we can differentiate

source/extensions/common/wasm/foreign.cc
TristonianJones: Errors won't necessarily terminate evaluation in CEL. The status_or let's you know whether there is a precondition failure in setting up the expression for evaluation, think a missing function binding. The error indicates whether there was a runtime evaluation error. Since you could combine CEL expression results and error values can be short-circuited away, it's important to preserve the distinction between an application configuration error and an expression evaluation error. We leverage this property extensively in computing an expression across multiple processes

jplevyak: How do pre-compilation errors fit in?  Are the execution environment errors all resolved at pre-compilation time or are they reported here as evaluate status (non-value) errors (e.g. because of late binding) ?

jplevyak: Stubby has the same issue with application errors, but they never really dealt with it in a consistent way.  This is definitely better than that.

TristonianJones: You know what, I was not looking very closely at the use of status_or before. This is what I get for trying to grok code on my phone. I was thinking this was the CreateExpression + Evaluate steps combined. A precondition failure normally happens on CreateExpression, whereas Error is an Evaluation concept. 

If Evaluation returns a non-okay status it should signal a CEL framework bug whereas Error is a user-space bug.

jplevyak:
LGTM, but please change the message to differentiate the two case.  Thanx!

source/extensions/common/wasm/foreign.cc
mandarjog: In this case it was missing activation key, so could be classified as either,  but I understand this is user error now. Thanks.

source/extensions/common/wasm/foreign.cc
mandarjog, line 175:
```suggestion
        ENVOY_LOG(debug, "expr_evaluate value error: {}", value.ErrorOrDie()->message());
```

mandarjog:
@jplevyak it wants me to add a test for the error case?

PR #521, opened 2020-05-13
Move control of the in-VM context lifecycle to the host specific code.

jplevyak:


jplevyak:
The failures at from the signature mismatch issue.  I'll see if I can move things around to work around.

source/extensions/common/wasm/context.cc
```diff
@@ -1364,6 +1391,13 @@ Http::FilterHeadersStatus Context::encode100ContinueHeaders(Http::ResponseHeader
 
 Http::FilterHeadersStatus Context::encodeHeaders(Http::ResponseHeaderMap& headers,
                                                  bool end_stream) {
+  if (!in_vm_context_created_) {
+    // If the request is invalid then onRequestHeaders() will not be called and neither will
+    // onCreate() then sendLocalReply be called which will call this function. In this case we
+    // need to call onCreate() so that the Context inside the VM is created before the
+    // onResponseHeaders() call.
+    onCreate(root_context_id_);
```
PiotrSikora, line 1399: I have an issue with this change, since it breaks normal HTTP request/response flow, and suddenly extensions will see a response to a request that it never received. If HTTP request is rejected by a lower layer (e.g. HTTP parser) or earlier extensions, and it never reaches our extension, then I don't see why HTTP response to that request should reach it.

As such, I think we should return `Http::FilterDataStatus::Continue`.

jplevyak: SGTM.  This matches the Envoy native filter behavior, but I agree that is odd.  That said, it is a weird case.  The filters normally run for the response back to downstream, and now we are saying that it will not happen for certain responses (i.e. malformed headers). Should I create an issue in envoyproxy/envoy asking what the correct behavior should be in general (not just for Wasm) ?

jplevyak: Done (code change)

source/extensions/common/wasm/context.cc
```diff
@@ -1244,6 +1256,9 @@ void Context::log(const Http::RequestHeaderMap* request_headers,
                   const Http::ResponseHeaderMap* response_headers,
                   const Http::ResponseTrailerMap* response_trailers,
                   const StreamInfo::StreamInfo& stream_info) {
+  if (!in_vm_context_created_) {
+    return;
```
PiotrSikora, line 1260: Should access logger still log rejected request (...even when it didn't process it?)

source/extensions/common/wasm/context.cc
PiotrSikora: Yeah, please open an issue upstream.

jplevyak: https://github.com/envoyproxy/envoy/issues/11447

source/extensions/common/wasm/context.cc
jplevyak: Humm... That interface is more independent.  However this is the onLog call which is called on a stream context.  There is a different type extension point which calls onLog on the Root context which is stateless.   That call will still  occur even if no other events have occurred. 

I would be tempted to return here because the user could have chosen the other extension point.  WDYT?

PiotrSikora: SGTM, although it's a bit confusing that we have 2 separate extension points to do the same thing.

jplevyak: I agree that it is a bit confusing, but this was a request from Mandar et al.   We can revisit it.

PR #524, opened 2020-05-15
Improve coverage of context.cc and remove duplicate code

jplevyak:
which is in proxy-wasm-cpp-host.


test/extensions/filters/http/wasm/test_data/queue_cpp.cc
```diff
@@ -43,7 +47,15 @@ void ExampleRootContext::onQueueReady(uint32_t token) {
     logInfo("onQueueReady");
   }
   std::unique_ptr<WasmData> data;
-  if (WasmResult::Ok == dequeueSharedQueue(token, &data)) {
+  if (dequeueSharedQueue(9999999 /* bad token */, &data) != WasmResult::NotFound) {
+    // Should not be called.
+    logWarn("onQueueReady bad token found");
```
PiotrSikora, line 52: Nit: I'd prefer you could turn those into positive tests, i.e.
```
if (dequeueSharedQueue(9999999 /* bad token */, &data) == WasmResult::NotFound) {
  logWarn("onQueueReady bad token not found");
}
```
so that we could verify that those things are actually happening.

(same in other tests)

PR #529, opened 2020-05-19
Wasm proto failopen

jplevyak:


jplevyak:
@mandarjog

api/envoy/extensions/wasm/v3/wasm.proto
```diff
@@ -65,12 +64,18 @@ message PluginConfig {
   // Filter/service configuration used to configure or reconfigure a plugin
   // (proxy_on_configuration).
   google.protobuf.Any configuration = 4;
+
+  // If there is a fatal error on the VM (e.g. exception, abort(), on_start or on_configure return false),
+  // then all plugins associated with the VM will either fail closed (by default), e.g. by returning an HTTP 503 error,
+  // or fail open (if 'fail_open' is set to true) by bypassing the filter. Note: when on_start or on_configure return false
```
mandarjog, line 70: `return false`

api/envoy/extensions/wasm/v3/wasm.proto
```diff
@@ -65,12 +64,18 @@ message PluginConfig {
   // Filter/service configuration used to configure or reconfigure a plugin
   // (proxy_on_configuration).
   google.protobuf.Any configuration = 4;
+
+  // If there is a fatal error on the VM (e.g. exception, abort(), on_start or on_configure return false),
```
mandarjog, line 68: Is the entire VM tainted or just the plugin? It was unexpected for me that all plugins within the VM are knocked out.

api/envoy/extensions/wasm/v3/wasm.proto
PiotrSikora: It says "return false", but "false" is outside of the default viewing window, so you need to scroll to the right to see it.

api/envoy/extensions/wasm/v3/wasm.proto
```diff
@@ -65,12 +64,18 @@ message PluginConfig {
...
+  // then all plugins associated with the VM will either fail closed (by default), e.g. by returning an HTTP 503 error,
+  // or fail open (if 'fail_open' is set to true) by bypassing the filter. Note: when on_start or on_configure return false
+  // during xDS updates the xDS configuration will be rejected and when on_start or on_configuration return false on initial
+  // startup the proxy will not start.
+  bool fail_open = 5;
```
PiotrSikora, line 73: Wasn't the consensus that we should always blackhole the traffic?

api/envoy/extensions/wasm/v3/wasm.proto
jplevyak: In general if the VM crashes (e.g. nullptr, abort()) then we cannot call into the VM safely and all plugins associated with that VM are tainted.

The question is whether or not we want to treat on_configure failures for a particular plugin specially.  I would say no, at least for now.   The reason is that it would be more complex to reason about and I don't think the complexity is warranted without a compelling use case.

api/envoy/extensions/wasm/v3/wasm.proto
jplevyak: I am not sure what this comment is referring to, could you clarify?

jplevyak: Sorry, I was working in a window without Piotr's comments.

api/envoy/extensions/wasm/v3/wasm.proto
jplevyak: Didn't see Piotr's comments.

api/envoy/extensions/wasm/v3/wasm.proto
jplevyak: I was under the impression that Mandar still wanted a fail closed/fail open option for telemetry.  This only effects non-startup non-xDS cases BTW.

mandarjog: If the telemetry extension fails to load, we should not blackhole traffic.
fail_open with default set to false is correct.

Issue #446, opened 2020-03-04
Indicate when a request has no body in WASM


gbrail:
C++ filters pass the "end-of-stream" flag to the HTTP header callbacks, and set it to true when a request or response has no body. However, we don't pass this flag via the WASM interface.

Without it, there's no way to know if there is going to be a body, so a filter that is supposed to parse or transform the body may never sees its body callbacks called and has no reliable way to know that this will happen.

If we're talking about changing the ABI anyway, then we should add the end_of_stream flag here too.

Otherwise, we can add some new ABI functions so that a filter can find out if it was already set.

gbrail:
@PiotrSikora : I saw some mention in your last PR about a "new ABI revision." If something like that is coming, would it make sense to try and address this issue first by adding the "end_of_stream" flag to the HTTP header callbacks?

PiotrSikora:
Good point, I'm not sure how we've missed that!

Will add it in the next iteration of the ABI. Thanks!

PR #532, opened 2020-05-21
Pass the "end_of_stream" flag to header callbacks.

gbrail:
Pass the end_of_stream flag to the onRequestHeaders and onResponseHeaders callbacks so that WASM proxies can later use them. (Right now they don't -- that will be addressed in another PR to proxy-wasm-cpp-host.)

This begins to address:

https://github.com/envoyproxy/envoy-wasm/issues/446

This won't build until
https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/21
is merged.


jplevyak:
proxy-wasm/proxy-wasm-cpp-host#21 has been merged

PR #538, opened 2020-05-28
Add support for onForeignFunction and proxy-specific extensions.

jplevyak:


test/extensions/common/wasm/test_data/test_context_cpp.cc
```diff
@@ -0,0 +1,46 @@
+// NOLINT(namespace-envoy)
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#ifndef NULL_PLUGIN
+#include "proxy_wasm_intrinsics.h"
+#include "source/extensions/common/wasm/ext/envoy_proxy_wasm_api.h"
```
Shikugawa, line 8: Can't we separate between this and wasm binary? I think that these proxy-specific extensions should be separated from envoy implementation like proxy-wasm. Do you have any idea to overcome this?

jplevyak: Do you mean we should create an alias in bazel so that these appear on a different directory such that we could create an Envoy specific SDK repo with the same header locations?  We could do that in a follow on PR.

Shikugawa: It means that whether we can define these headers as single github repository. So to speak, whether we can create a repository like `envoyproxy/envoy-proxy-wasm-cpp-sdk`. This header implementation seems to be completely separated from upstream envoy so that we can achieve this. It will be more understandable for filter developers.

source/extensions/common/wasm/context.h
```diff
@@ -398,6 +414,7 @@ class Context : public proxy_wasm::ContextBase,
   ::Envoy::Buffer::Instance* network_upstream_data_buffer_{};
 
   // HTTP filter state.
+  bool http_request_started_ = false; // When decodeHeaders() is called the rquest is "started".
```
PiotrSikora, line 417: `s/rquest/request/`

bazel/external/wee8.genrule_cmd
```diff
@@ -54,10 +54,12 @@ fi
 # Debug/release build.
 if [[ $(COMPILATION_MODE) == "dbg" && $${ENVOY_UBSAN_VPTR-} != "1" && $${ENVOY_MSAN-} != "1" && $${ENVOY_TSAN-} != "1" ]]; then
   WEE8_BUILD_ARGS+=" is_debug=true"
+  WEE8_BUILD_ARGS+=" v8_symbol_level=2"
   WEE8_BUILD_ARGS+=" symbol_level=2"
```
PiotrSikora, line 58: We only need `v8_symbol_level`.

test/extensions/common/wasm/test_data/test_context_cpp.cc
jplevyak: We should do that once we snap a stable LTS version 1 ABI/SDK. Let me open an issue in envoyproxy/envoy: https://github.com/envoyproxy/envoy/issues/11755

source/extensions/common/wasm/context.h
jplevyak: Done.

bazel/external/wee8.genrule_cmd
jplevyak: Done

test/extensions/common/wasm/test_data/test_context_cpp.cc
jplevyak: We will need a new repo under envoyproxy so we need Matt to create that.

jplevyak:
Ping.  This is the other half of https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/23.

PR #547, opened 2020-06-12
fix early deletion of wasm state schema

kyessenov:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -950,8 +950,12 @@ WasmResult Context::setProperty(absl::string_view path, absl::string_view value)
 
 WasmResult Context::declareProperty(absl::string_view path,
                                     std::unique_ptr<const WasmStatePrototype> state_prototype) {
-  state_prototypes_[path] = std::move(state_prototype);
-  return WasmResult::Ok;
+  // Do not delete existing schema since it can be referenced by state objects.
+  if (state_prototypes_.find(path) == state_prototypes_.end()) {
+    state_prototypes_[path] = std::move(state_prototype);
+    return WasmResult::Ok;
+  }
+  return WasmResult::BadArgument;
 }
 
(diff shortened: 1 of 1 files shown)


kyessenov:
The return result changed, but it's not checked by metadata exchange plugin at the moment. Behaviourally it's the same as long as double registration happens to be identical.

source/extensions/common/wasm/context.cc
```diff
@@ -950,8 +950,12 @@ WasmResult Context::setProperty(absl::string_view path, absl::string_view value)
 
 WasmResult Context::declareProperty(absl::string_view path,
                                     std::unique_ptr<const WasmStatePrototype> state_prototype) {
-  state_prototypes_[path] = std::move(state_prototype);
-  return WasmResult::Ok;
+  // Do not delete existing schema since it can be referenced by state objects.
+  if (state_prototypes_.find(path) == state_prototypes_.end()) {
+    state_prototypes_[path] = std::move(state_prototype);
+    return WasmResult::Ok;
+  }
```
mandarjog, line 957: It should be bad argument only if the new and the old registration is different, right?

kyessenov: That would require equality for the wasm prototypes. In general, it's not clear to me what to do if this method fails in the client. So the client should just log and ignore.

mandarjog:
@PiotrSikora @jplevyak We need this in 1.6...

source/extensions/common/wasm/context.cc
kyessenov: I'd like to get rid of protobuf as an argument eventually. The return result is currently ignored.

jplevyak: SGTM.

PR #548, opened 2020-06-16
Add support for closing of connections in the ABI/SDK.

jplevyak:


source/extensions/common/wasm/context.cc
```diff
@@ -1305,6 +1305,32 @@ WasmResult Context::continueStream(StreamType stream_type) {
...
+    return WasmResult::Ok;
+  case WasmStreamType::Response:
+    if (encoder_callbacks_) {
+      encoder_callbacks_->resetStream();
+    }
+    return WasmResult::Ok;
+  case WasmStreamType::Downstream:
+    if (network_read_filter_callbacks_) {
+      network_read_filter_callbacks_->connection().close(
+          Envoy::Network::ConnectionCloseType::FlushWrite);
```
kyessenov, line 1323: Why always FlushWrite?

jplevyak: @PiotrSikora and I had a long talk about this.  We certainly do not want to wait, so it was either WriteFlush or NoFlush.  We decided on WriteFlush although to be honest I could go either way.

jplevyak:
Ping.  The support for this has already gone into the -sdk and -host code.

jplevyak:
Ping, could I get a review, I have another PR which is being held up by this one.

source/extensions/common/wasm/context.cc
```diff
@@ -1305,6 +1305,32 @@ WasmResult Context::continueStream(StreamType stream_type) {
...
+WasmResult Context::closeStream(WasmStreamType stream_type) {
+  switch (stream_type) {
+  case WasmStreamType::Request:
+    if (decoder_callbacks_) {
+      decoder_callbacks_->resetStream();
+    }
+    return WasmResult::Ok;
+  case WasmStreamType::Response:
+    if (encoder_callbacks_) {
+      encoder_callbacks_->resetStream();
```
PiotrSikora, line 1317: Nit: You probably should reset both streams in both cases.

jplevyak: That is a very 1:1 mindset. A more general model is a transformer multiplexer model which is not strictly 1:1.  When we merge with the other WASI proposal the model will have to become more general so best not to bake a restrictive model in now.

PiotrSikora: This is/was about envoy-wasm implementation, where 1:1 relation is baked, and not WASI standard.

PR #557, opened 2020-06-25
Handle Wasm errors by failing closed or open.

jplevyak:


PiotrSikora:
cc @mandarjog

source/extensions/access_loggers/wasm/config.cc
```diff
@@ -32,25 +32,39 @@ WasmAccessLogFactory::createAccessLogInstance(const Protobuf::Message& proto_con
...
     // NB: the Slot set() call doesn't complete inline, so all arguments must outlive this call.
-    tls_slot->set([base_wasm, plugin](Event::Dispatcher& dispatcher) {
-      return std::static_pointer_cast<ThreadLocal::ThreadLocalObject>(
-          Common::Wasm::getOrCreateThreadLocalWasm(base_wasm, plugin, dispatcher));
-    });
+    tls_slot->set(
+        [base_wasm,
+         plugin](Event::Dispatcher& dispatcher) -> std::shared_ptr<ThreadLocal::ThreadLocalObject> {
+          if (!base_wasm) {
+            // Should this be a RELEASE_ASSERT? There is no way to prevent the connection at this
```
PiotrSikora, line 47: Nope. `RELEASE_ASSERT` should be used for code failures, not for errors triggered by invalid configuration.

source/extensions/common/wasm/wasm.cc
```diff
@@ -134,6 +134,8 @@ Wasm::Wasm(WasmHandleSharedPtr base_wasm_handle, Event::Dispatcher& dispatcher)
   ENVOY_LOG(debug, "Thread-Local Wasm created {} now active", active_wasms);
 }
 
+void Wasm::error(absl::string_view message) { ENVOY_LOG(trace, "Wasm VM failed {}", message); }
```
PiotrSikora, line 137: This should be logged at a level higher than `trace`.

test/extensions/bootstrap/wasm/config_test.cc
```diff
@@ -84,15 +84,16 @@ TEST_P(WasmFactoryTest, CreateWasmFromWasmPerThread) {
...
       absl::StrCat("envoy.wasm.runtime.", GetParam()));
   config.mutable_config()->mutable_vm_config()->mutable_code()->mutable_local()->set_filename(
       TestEnvironment::substitute(
           "{{ test_rundir }}/test/extensions/bootstrap/wasm/test_data/missing_cpp.wasm"));
   config.set_singleton(true);
 
-  EXPECT_THROW_WITH_REGEX(initializeWithConfig(config), Extensions::Common::Wasm::WasmException,
-                          "Failed to load Wasm module due to a missing import: env.missing");
+  EXPECT_THROW_WITH_MESSAGE(initializeWithConfig(config), Extensions::Common::Wasm::WasmException,
+                            "Unable to create Wasm service test");
```
PiotrSikora, line 96: This doesn't test what it is supposed to test anymore. What happened to the original error message? The cpp-host part still emits it.

test/extensions/bootstrap/wasm/config_test.cc
```diff
@@ -110,19 +111,20 @@ TEST_P(WasmFactoryTest, UnspecifiedRuntime) {
...
   config.mutable_config()->mutable_vm_config()->set_runtime("envoy.wasm.runtime.invalid");
   config.mutable_config()->mutable_vm_config()->mutable_code()->mutable_local()->set_filename(
       TestEnvironment::substitute(
           "{{ test_rundir }}/test/extensions/bootstrap/wasm/test_data/start_cpp.wasm"));
   config.set_singleton(true);
 
   EXPECT_THROW_WITH_MESSAGE(initializeWithConfig(config), Extensions::Common::Wasm::WasmException,
-                            "Failed to create Wasm VM using envoy.wasm.runtime.invalid runtime. "
-                            "Envoy was compiled without support for it.");
+                            "Unable to create Wasm service test");
```
PiotrSikora, line 122: (same here)

test/extensions/bootstrap/wasm/wasm_test.cc
```diff
@@ -130,61 +135,44 @@ TEST_P(WasmTest, BadSignature) {
...
   if (GetParam() == "v8") {
-    EXPECT_THROW_WITH_MESSAGE(
-        context->onLog(), Extensions::Common::Wasm::WasmException,
-        "Function: proxy_on_log failed: Uncaught RuntimeError: divide by zero");
-  } else if (GetParam() == "wavm") {
-    EXPECT_THROW_WITH_REGEX(context->onLog(), Extensions::Common::Wasm::WasmException,
-                            "Function: proxy_on_log failed: wavm.integerDivideByZeroOrOverflow.*");
-  } else {
-    ASSERT_FALSE(true); // Neither of the above was matched.
+    // We are not throwing out of V8 and V8 does not unwind the stack, but continues execution.
```
PiotrSikora, line 163: Where did that change in behavior come from? This test was catching it before (see: `Uncaught RuntimeError: divide by zero` message generated by V8).

test/extensions/bootstrap/wasm/config_test.cc
jplevyak: The original error message is coming out in a log statement.  Is there any way of capturing those log statements in tests without having to resort to intrusive changes e.g. adding code to intercept the logs manually ?  Alternatively I can stored the original log message as say fail_message_ and pass it back via the exception.

test/extensions/bootstrap/wasm/wasm_test.cc
jplevyak: When failures occur we no longer throw an exception because those exceptions where not caught and resulted in crashes.  Now we just set a failed flag (as V8 does).   We could throw in the special case of divide by zero and wrap everything in a try block, but I didn't go that route. As you are aware there is resistance to using exceptions in some quarters and it wasn't clear that using them in this one case was important enough to justify it.

test/extensions/bootstrap/wasm/config_test.cc
jplevyak: See above.

source/extensions/common/wasm/wasm.cc
jplevyak: Changed to "error".

source/extensions/access_loggers/wasm/config.cc
jplevyak: Updated commments.

test/extensions/bootstrap/wasm/config_test.cc
PiotrSikora: I guess it's fine if it's emitted in the logs. It's a bit unfortunate that we emit 2 separate messages, and normally I'd ask to retain and resurface the root source of failure, but I have no idea what upstream's take on that will be, so let's leave it as-is, and let them bikeshed.

test/extensions/bootstrap/wasm/wasm_test.cc
PiotrSikora: Sorry, let me rephrase. I don't care about exception vs failure flag.

What I don't like here is that the execution continues, which results in undefined behavior.

Unless I'm missing something, V8 is still going to trap on div by zero, instead of returning the result, and previously (i.e. before https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/29) we'd `abort()` in that case, but now we're only setting the failure flag. Shouldn't that also stop further execution?

jplevyak: If I call abort in the handler then it will abort Envoy itself.  I don't know how to tell V8 to stop execution and return an error, but perhaps you do ?  I had assumed that if I just return from the trap handler then it would abort execution, but it is continuing.

jplevyak: Sorry for the noise.  Something is wrong with github.  The above code is not what is actually in the branch at commit a2040d57bbb9c35ca57dfc5c5021cbfa3ea4646b.   Please pull my branch and take a look at this code.  It does in fact do was you are saying.  What you are seeing seems to be some intermediate state.

PR #565, opened 2020-07-08
Remove rust API and .wasm files.

jplevyak:


jplevyak:
We need the rust bazel build system before we can remove these.

PR #574, opened 2020-07-13
Add Null VM tests and merge  some tests to decrease compile time.

jplevyak:
This depends on https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/31


jplevyak:
ping

test/extensions/bootstrap/wasm/config_test.cc
```diff
@@ -48,113 +67,69 @@ class WasmFactoryTest : public testing::TestWithParam<std::string> {
 };
 
 INSTANTIATE_TEST_SUITE_P(Runtimes, WasmFactoryTest,
-                         testing::Values("v8"
+                         testing::Values("v8",
 #if defined(ENVOY_WASM_WAVM)
```
lizan, line 71: FYI this pattern (`#if` inside macro expansion) breaks MSVC

PiotrSikora: @jplevyak Should we just nuke WAVM support?

jplevyak: Yes.  I am going to do that as soon as I merge this.

PR #580, opened 2020-07-16
Remove WAVM support (for now).

jplevyak:


bazel/BUILD
```diff
@@ -253,11 +253,6 @@ config_setting(
     values = {"define": "wasm=enabled"},
 )
```
PiotrSikora, line 254: I don't think we need `wasm=enabled` either, since V8 is enabled by default.

jplevyak: Done.

PR #582, opened 2020-07-17
Check for failed Wasm in on_resolved_dns().

jplevyak:
Log a critical error if a Wasm service fails to load
and it is set to fail closed.


jplevyak:
Ping.  Could I get a review ?

jplevyak:
Ping

PR #589, opened 2020-07-22
Add support for Wasm distribution failure mode stats.

jplevyak:


test/extensions/bootstrap/wasm/wasm_speed_test.cc
```diff
@@ -59,7 +59,7 @@ static void BM_WasmSimpleCallSpeedTest(benchmark::State& state, std::string test
   auto vm_key = "";
   auto plugin_configuration = "";
   auto plugin = std::make_shared<Extensions::Common::Wasm::Plugin>(
-      name, root_id, vm_id, plugin_configuration, false,
+      name, root_id, vm_id, vm, plugin_configuration, false,
```
PiotrSikora, line 62: Nit: might be worth changing `vm` to `runtime` to be consistent across the codebase.

source/extensions/common/wasm/wasm_extension.cc
```diff
@@ -62,6 +70,8 @@ WasmHandleExtensionCloneFactory EnvoyWasm::wasmCloneFactory() {
 
 void EnvoyWasm::onEvent(WasmEvent event, const PluginSharedPtr&) {
   switch (event) {
+  default:
+    break;
```
PiotrSikora, line 74: Nit: move to the end?

source/extensions/common/wasm/wasm.cc
```diff
@@ -272,6 +277,30 @@ getCloneFactory(WasmExtension* wasm_extension, Event::Dispatcher& dispatcher,
...
+    return WasmEvent::MissingFunction;
+  case FailState::UnableToInitializeCode:
+    return WasmEvent::UnableToInitializeCode;
+  case FailState::StartFailed:
+    return WasmEvent::StartFailed;
+  case FailState::ConfigureFailed:
+    return WasmEvent::ConfigureFailed;
+  case FailState::RuntimeError:
+    return WasmEvent::RuntimeError;
+  }
```
PiotrSikora, line 301: Nit: add `default` case for older compilers (I think we still need that?)

jplevyak: There are other exhaustive cases, e.g. Context::log.  Added NOT_IMPLEMENTED_GCOVR_EXCL_LINE which seems to be the solution to this.

source/extensions/common/wasm/wasm_extension.cc
jplevyak: Done.

test/extensions/bootstrap/wasm/wasm_speed_test.cc
jplevyak: Done.

source/extensions/common/wasm/wasm.cc
PiotrSikora: `NOT_IMPLEMENTED_GCOVR_EXCL_LINE` should be added as the `default` case, not outside of the `switch`.

jplevyak: This is how it is done in context.cc.  This enables the compiler which does correctly support exhaustive cases to report those errors.  Which compiler do you think does not support this?  Given that Envoy is now on c++17, I don't think those compilers are supported any longer.

PiotrSikora: I was thinking about GCC (see: #499). If you want to use it outside of the `default` case, then `NOT_REACHED_GCOVR_EXCL_LINE` is more appropriate than `NOT_IMPLEMENTED_GCOVR_EXCL_LINE`.

I'm assuming that C++ standard doesn't change compiler's behavior, since the codebase is filled with those.

jplevyak: Done. PTAL

PR #592, opened 2020-07-28
Add Wasm config option to nack_on_code_cache_miss.

jplevyak:

### DIFF
--- a/api/envoy/extensions/wasm/v3/wasm.proto
+++ b/api/envoy/extensions/wasm/v3/wasm.proto
@@ -18,7 +18,7 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // [#protodoc-title: Wasm service]
 
 // Configuration for a Wasm VM.
-// [#next-free-field: 6]
+// [#next-free-field: 7]
 message VmConfig {
   // An ID which will be used along with a hash of the wasm code (or the name of the registered Null
   // VM plugin) to determine which VM will be used for the plugin. All plugins which use the same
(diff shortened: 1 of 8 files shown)


jplevyak:
@mandarjog @bianpengyuan

api/envoy/extensions/wasm/v3/wasm.proto
```diff
@@ -40,6 +40,11 @@ message VmConfig {
   // Warning: this should only be enable for trusted sources as the precompiled code is not
   // verified.
   bool allow_precompiled = 5;
+
+  // If true and the code needs to be remotely fetched and it is not in the cache then NACK the configuration
+  // update and do a background fetch to fill the cache, otherwise fetch the code asynchronously and enter
+  // warming state.
+  bool nack_on_code_cache_miss = 6;
```
mandarjog, line 47: Can this just be filter config in wasm proto? 2 filters using same vm Id can have different values here and it would be confusing.

jplevyak: This is in the vm config.  Since that is the place where the remote fetch URL is, it is the place where we need to specifiy the behavior if the URL fetch fails.

mandarjog: Ok, so it will only apply where it is specified. Which means if 2 configurations specify same vm id but different NACk behaviour it will work as expected.

jplevyak: The first one will win in this case unlesss we add that bit into the vm_key.   A single Base Wasm is created for each vm_key by the first plugin to require that vm_key and applies to all plugins with that vm_key.  This is one of the reasons I was reluctant to add this feature.  It adds a lot of conceptual complexity and will require educating the user and all to work around limitations with xDS which because it has "eventual consistency" and no way to report async status is itself conceptually complex.  We are just layering problems to work around other problems.

Oh and adding that bit to the vm_key is still complex because it isn't intuitive at all that it would cause the VMs to split.

PR #596, opened 2020-08-02
Support multiple versions of the Proxy-Wasm ABI.

PiotrSikora:


PiotrSikora:
(Note: `asan` integration tests with external repositories are broken on `master`)

jplevyak:
I would have expected a switch somewhere, although perhaps we don't have any incompatible changes at this point.  If that is the case, please add a comment in the appropriate place (e.g. the docs or where the version functions are registered).

PiotrSikora:
> I would have expected a switch somewhere, although perhaps we don't have any incompatible changes at this point. If that is the case, please add a comment in the appropriate place (e.g. the docs or where the version functions are registered).

All the switching is happening in Proxy-Wasm C++ Host implementation (see: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/42).

Note, that because we register/expose host functions before linking, we don't know the exported ABI version marker at that point, so we expose host functions for both 0.1.0 and 0.2.0 version of the ABI (since they are not conflicting), and the switching is only happening for the callbacks.

We could expose host functions only for the requested ABI version, but that would require some invasive rewrites to the code that I'm not willing to make just before the release.

PR #597, opened 2020-08-02
Defer sendLocalReply() to prevent reentrant calls.

PiotrSikora:
This was previously fixed in #432, but got somehow lost.

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1377,8 +1377,12 @@ WasmResult Context::sendLocalResponse(uint32_t response_code, absl::string_view 
     }
   };
   if (decoder_callbacks_) {
-    decoder_callbacks_->sendLocalReply(static_cast<Envoy::Http::Code>(response_code), body_text,
-                                       modify_headers, grpc_status, details);
+    wasm()->addAfterVmCallAction([this, response_code, body_text = std::string(body_text),
+                                  modify_headers = std::move(modify_headers), grpc_status,
+                                  details = std::string(details)] {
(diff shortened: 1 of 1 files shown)


source/extensions/common/wasm/context.cc
```diff
@@ -1377,8 +1377,12 @@ WasmResult Context::sendLocalResponse(uint32_t response_code, absl::string_view
...
   };
   if (decoder_callbacks_) {
-    decoder_callbacks_->sendLocalReply(static_cast<Envoy::Http::Code>(response_code), body_text,
-                                       modify_headers, grpc_status, details);
+    wasm()->addAfterVmCallAction([this, response_code, body_text = std::string(body_text),
+                                  modify_headers = std::move(modify_headers), grpc_status,
+                                  details = std::string(details)] {
+      decoder_callbacks_->sendLocalReply(static_cast<Envoy::Http::Code>(response_code), body_text,
+                                         modify_headers, grpc_status, details);
+    });
```
PiotrSikora, line 1385: Let me know if `if (decoder_callbacks_) { ... }` should be inside the deferred lambda.

jplevyak: This is a bit subtle because proxy_on_delete does call DeferAfterCallActions so in theory it could call this and the Context in the VM would be invalid, but because it only gets called after the connections have drained, the call to sendLocalReply will fail.  Net net, this is safe, but I would add a N.B. since the why is not obvious.

PiotrSikora: Added comment, but it doesn't look like this is taken into account in other places that defer calls.

PR #605, opened 2020-08-05
Implement proxy_get_log_level

gbrail:
This hooks up the proxy runtime to the support that's already been added to the host and SDK libraries.

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1110,6 +1110,12 @@ WasmResult Context::log(uint32_t level, absl::string_view message) {
   NOT_REACHED_GCOVR_EXCL_LINE;
 }
 
+uint32_t Context::getLogLevel() {
+  // Like the "log" call above, assume that spdlog level as an int
+  // matches the enum in the SDK
+  return static_cast<uint32_t>(ENVOY_LOGGER().level());
+}
+
 //
 // Calls into the Wasm code.
 //
--- a/bazel/repository_locations.bzl
+++ b/bazel/repository_locations.bzl
@@ -513,9 +513,9 @@ DEPENDENCY_REPOSITORIES = dict(
         cpe = "N/A",
     ),
     proxy_wasm_cpp_host = dict(
-        sha256 = "8932d72e321c7c527794d12e1864ddab95276b77bebb0b2c2f6fbc49b2b4b39a",
-        strip_prefix = "proxy-wasm-cpp-host-5afe9bdc9d3dc833522332b5c0b87e1339fd2da2",
-        urls = ["https://github.com/proxy-wasm/proxy-wasm-cpp-host/archive/5afe9bdc9d3dc833522332b5c0b87e1339fd2da2.tar.gz"],
+        sha256 = "1cc5334ce433507371d7ccb58c87e07ca6bd1582a75a5dafe02a89a54106db98",
+        strip_prefix = "proxy-wasm-cpp-host-4598dc3d0d3b731be1dcccb400732c72bebb1380",
+        urls = ["https://github.com/proxy-wasm/proxy-wasm-cpp-host/archive/4598dc3d0d3b731be1dcccb400732c72bebb1380.tar.gz"],
         use_category = ["dataplane"],
         cpe = "N/A",
     ),
--- a/source/extensions/common/wasm/context.h
+++ b/source/extensions/common/wasm/context.h
@@ -163,6 +163,7 @@ public:
            const Http::ResponseHeaderMap* response_headers,
            const Http::ResponseTrailerMap* response_trailers,
            const StreamInfo::StreamInfo& stream_info) override;
+  uint32_t getLogLevel() override;
 
   // Network::ConnectionCallbacks
   void onEvent(Network::ConnectionEvent event) override;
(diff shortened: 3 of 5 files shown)


source/extensions/common/wasm/context.cc
```diff
@@ -1110,6 +1110,12 @@ WasmResult Context::log(uint32_t level, absl::string_view message) {
   NOT_REACHED_GCOVR_EXCL_LINE;
 }
 
+uint32_t Context::getLogLevel() {
+  // Like the "log" call above, assume that spdlog level as an int
+  // matches the enum in the SDK
+  return static_cast<uint32_t>(ENVOY_LOGGER().level());
+}
```
PiotrSikora, line 1117: It might be a bit too late to ask this question, but how do you intend to use this?

Log level can change at runtime (at least in Envoy), so either you have to actively check the log level before each log attempt (expensive) or you're facing the possibility of not producing the log entries.

Wouldn't a `proxy_on_log_level_change` callback be more useful?

gbrail:
Yeah, I agree that calling it on every logger call would be a bummer for
performance. I was instead planning to use it to set the log level for my
filter. Without it, every "trace" and "debug" call in my filter code has to
generate a log message and make the ABI call to have it be logged, where
it's then discarded.

I agree that a proxy_on_log_level_change callback would be ideal (although
we need a way to initialize it, which could still be this call).

I'm not sure where to hook that in Envoy, though.

source/extensions/common/wasm/context.cc
jplevyak: I agree, but this is a start.

jplevyak:
Logger::setLevel, but then you need a mechanism to register interest in those changes and then the notifications have to be delivered to each thread and the Wasm VM iterated through.

PiotrSikora:
Coverage test fails:
```
test trace logging
[2020-08-06 03:19:17.986][25904][trace][wasm] [source/extensions/common/wasm/context.cc:1089] wasm log : test trace logging
test debug logging
[2020-08-06 03:19:17.986][25904][debug][wasm] [source/extensions/common/wasm/context.cc:1092] wasm log : test debug logging
test warn logging
[2020-08-06 03:19:17.986][25904][warning][wasm] [source/extensions/common/wasm/context.cc:1098] wasm log : test warn logging
test error logging
[2020-08-06 03:19:17.986][25904][error][wasm] [source/extensions/common/wasm/context.cc:1101] wasm log : test error logging
log level is 0
unknown file: Failure

Unexpected mock function call - returning directly.
    Function call: log_(2, "log level is 0")
Google Mock tried the following 8 expectations, but none matched:

test/extensions/common/wasm/wasm_test.cc:108: tried expectation #0: EXPECT_CALL(*root_context, log_(spdlog::level::info, Eq("on_configuration configure-test")))...
  Expected arg #1: is equal to 0x2b3da1 pointing to "on_configuration configure-test"
           Actual: "log level is 0"
         Expected: to be called once
           Actual: never called - unsatisfied and active
test/extensions/common/wasm/wasm_test.cc:109: tried expectation #1: EXPECT_CALL(*root_context, log_(spdlog::level::trace, Eq("test trace logging")))...
  Expected arg #0: is equal to 0
           Actual: 2
  Expected arg #1: is equal to 0x2b3dc1 pointing to "test trace logging"
           Actual: "log level is 0"
         Expected: to be called once
           Actual: called once - saturated and active
test/extensions/common/wasm/wasm_test.cc:110: tried expectation #2: EXPECT_CALL(*root_context, log_(spdlog::level::debug, Eq("test debug logging")))...
  Expected arg #0: is equal to 1
           Actual: 2
  Expected arg #1: is equal to 0x273c0a pointing to "test debug logging"
           Actual: "log level is 0"
         Expected: to be called once
           Actual: called once - saturated and active
test/extensions/common/wasm/wasm_test.cc:111: tried expectation #3: EXPECT_CALL(*root_context, log_(spdlog::level::warn, Eq("test warn logging")))...
  Expected arg #0: is equal to 3
           Actual: 2
  Expected arg #1: is equal to 0x26d2a4 pointing to "test warn logging"
           Actual: "log level is 0"
         Expected: to be called once
           Actual: called once - saturated and active
test/extensions/common/wasm/wasm_test.cc:112: tried expectation #4: EXPECT_CALL(*root_context, log_(spdlog::level::err, Eq("test error logging")))...
  Expected arg #0: is equal to 4
           Actual: 2
  Expected arg #1: is equal to 0x2406ee pointing to "test error logging"
           Actual: "log level is 0"
         Expected: to be called once
           Actual: called once - saturated and active
test/extensions/common/wasm/wasm_test.cc:113: tried expectation #5: EXPECT_CALL(*root_context, log_(spdlog::level::info, Eq("log level is 4")))...
  Expected arg #1: is equal to 0x259346 pointing to "log level is 4"
           Actual: "log level is 0"
         Expected: to be called once
           Actual: never called - unsatisfied and active
test/extensions/common/wasm/wasm_test.cc:114: tried expectation #6: EXPECT_CALL(*root_context, log_(spdlog::level::info, Eq("on_done logging")))...
  Expected arg #1: is equal to 0x279f8f pointing to "on_done logging"
           Actual: "log level is 0"
         Expected: to be called once
           Actual: never called - unsatisfied and active
test/extensions/common/wasm/wasm_test.cc:115: tried expectation #7: EXPECT_CALL(*root_context, log_(spdlog::level::info, Eq("on_delete logging")))...
  Expected arg #1: is equal to 0x2ba411 pointing to "on_delete logging"
           Actual: "log level is 0"
         Expected: to be called once
           Actual: never called - unsatisfied and active
```

Coverage tests are not hermetic, so another test must have changed the global log level. I think you need to set it explicitly at the start of the test, otherwise it's going to be flaky.

jplevyak:
Try setting the log level default in the test fixture.

gbrail:
Thanks -- I tried just setting the log level around the specific test and
it caused other problems so let me try that.

gbrail:
Fixing the tests is pretty easy, but to rebase this against master we need to fix the symbol I forgot to add to proxy_wasm_cpp_host:

https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/51

PR #606, opened 2020-08-05
Fix header corruption in sendLocalResponse

gbrail:
Now that WASM calls sendLocalReply as a deferred call, the
collection of string_views that it uses to set headers on the
new response points to invalid data. We have to make a copy of all
the new headers before deferring the call.

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1468,12 +1468,22 @@ WasmResult Context::sendLocalResponse(uint32_t response_code, absl::string_view 
                                       Pairs additional_headers, uint32_t grpc_status,
                                       absl::string_view details) {
 
-  auto modify_headers = [additional_headers](Http::HeaderMap& headers) {
-    for (auto& p : additional_headers) {
-      const Http::LowerCaseString lower_key{std::string(p.first)};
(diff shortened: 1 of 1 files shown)


source/extensions/common/wasm/context.cc
```diff
@@ -1468,12 +1468,22 @@ WasmResult Context::sendLocalResponse(uint32_t response_code, absl::string_view
                                       Pairs additional_headers, uint32_t grpc_status,
                                       absl::string_view details) {
 
-  auto modify_headers = [additional_headers](Http::HeaderMap& headers) {
-    for (auto& p : additional_headers) {
-      const Http::LowerCaseString lower_key{std::string(p.first)};
-      headers.addCopy(lower_key, std::string(p.second));
+
```
PiotrSikora, line 1471: Nit: 2 empty lines, which breaks the format check.

gbrail: Weird -- I must have the wrong version of clang-format or something because it passes for me. But it's easy to fix.

PiotrSikora: Thanks!

jplevyak:
FYI: envoy recently moved to clang-format-10 which is subtly different.

PR #609, opened 2020-08-07
Fix compatibility with Proxy-Wasm ABI v0.1.0.

PiotrSikora:

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -681,7 +681,12 @@ WasmResult Context::getHeaderMapValue(WasmHeaderMapType type, absl::string_view 
   const Http::LowerCaseString lower_key{std::string(key)};
   auto entry = map->get(lower_key);
   if (!entry) {
-    return WasmResult::NotFound;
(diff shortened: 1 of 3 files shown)


PiotrSikora:
Actually, this should probably include https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/52 when it's merged.

source/extensions/common/wasm/context.cc
```diff
@@ -1136,7 +1141,13 @@ bool Context::validateConfiguration(absl::string_view configuration,
   return result;
 }
 
-absl::string_view Context::getConfiguration() { return configuration_; }
+absl::string_view Context::getConfiguration() {
+  if (plugin_) {
```
jplevyak, line 1145: This breaks validateConfiguration.

PiotrSikora: How come?

jplevyak: Then let's just remove the configuration_ member.

PiotrSikora: Done.

PR #611, opened 2020-08-07
Add tests written using Proxy-Wasm Rust SDK.

PiotrSikora:


test/extensions/filters/network/wasm/wasm_filter_test.cc
```diff
@@ -76,9 +78,10 @@ TEST_P(WasmNetworkFilterTest, BadCode) {
 // Test happy path.
 TEST_P(WasmNetworkFilterTest, HappyPath) {
   const std::string code =
-      GetParam() != "null"
-          ? TestEnvironment::readFileToStringForTest(TestEnvironment::substitute(
-                "{{ test_rundir }}/test/extensions/filters/network/wasm/test_data/test_cpp.wasm"))
+      std::get<0>(GetParam()) != "null"
+          ? TestEnvironment::readFileToStringForTest(TestEnvironment::substitute(absl::StrCat(
+                "{{ test_rundir }}/test/extensions/filters/network/wasm/test_data/",
+                std::get<1>(GetParam()) == "cpp" ? "test_cpp" : "logging_rust", ".wasm")))
```
lizan, line 84: just rename logging_rust to test_rust, or the oppostie( test_cpp -> logging_cpp)?

PiotrSikora: I could, but I expect that those will diverge over time (see how this is handled in the HTTP filters, where all C++ tests were moved into `test_cpp.wasm` monolith to avoid long build times with Emscripten under Bazel, but we don't have such issue with Rust - at least not yet - so those tests are self-contained and in separate files).

But if you feel strongly about it, I can rename it.

lizan: Nope this isn't a blocker. I think I addressed emscripten long build time, but if that's not the case I'll look it again.

PiotrSikora: I think you did, but since we have only one C++ file per test now, it's hard to tell.

In any case, if this is not a blocker, then could you approve this PR as-is to avoid another CI cycle?

PR #612, opened 2020-08-10
ci: fix spell format

lizan:
This should let the branch pass all upstream format check.

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -217,7 +217,7 @@ void Context::onStatsUpdate(Envoy::Stats::MetricSnapshot& snapshot) {
   //    name
   //    8 byte alignment padding
   //    8 bytes of absolute value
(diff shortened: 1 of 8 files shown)


jplevyak:
I have very similar changes in the upstream PR, so we could wait for that.  @lizan what do you think the chances are to get that in in the not too distant future?  The main issues now seem to be ARM and Windows.

source/extensions/common/wasm/context.cc
```diff
@@ -554,7 +554,7 @@ WasmResult Context::getProperty(absl::string_view path, std::string* result) {
     start = end + 1;
 
     if (first) {
-      // top-level ident
+      // top-level indent
```
jplevyak, line 557: This should be "identifier".

lizan:
@jplevyak I think it will take a while so let's merge this to avoid future break (modified CI check)

jplevyak:
Agreed. Let's merge this and I"ll merge it into the upstream PR.

PR #613, opened 2020-08-10
Porting to build Rust wasm test datas with Bazel

Shikugawa:
Signed-off-by: Shikugawa <Shikugawa@gmail.com>

Commit Message: Port to enable building Rust wasm test datas with Bazel with some of cargo-raze supporting scripts.
Additional Description:
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


test/extensions/filters/http/wasm/wasm_filter_test.cc
```diff
@@ -77,14 +80,18 @@ class WasmHttpFilterTest : public Common::Wasm::WasmHttpFilterTestBase<
...
+      code = "HttpWasmTestCpp";
+    } else {
+      if (std::get<1>(GetParam()) == "cpp") {
+        code = TestEnvironment::readFileToStringForTest(TestEnvironment::substitute(
+            absl::StrCat("{{ test_rundir }}/", kWasmCppDir, "test_cpp.wasm")));
+      } else {
+        code = TestEnvironment::readFileToStringForTest(TestEnvironment::substitute(
+            absl::StrCat("{{ test_rundir }}/", kWasmRustDir, vm_configuration + "_rust.wasm")));
+      }
+    }
```
PiotrSikora, line 94: Why did you move all those test files out of their current location? It seems like an unnecessary change that requires extra switches.

Shikugawa: This is because of the limitation of cargo-raze. We should generate build dependencies per Rust projects. But it will cause more complexity.

jplevyak:
Do we want this as part of the current upstream PR?

jplevyak:
Looks like @htuch wants this in the upstream PR. @PiotrSikora @Shikugawa

bazel/external/cargo/remote/ahash-0.3.8.BUILD
```diff
@@ -0,0 +1,46 @@
+"""
```
htuch, line 1: Is the plan to check all of this into the main Envoy repo? Seems kind of terrifying from a maintenance perspective, even if automatically generated.

Shikugawa: Exactly. But maintaining these automatically generated is not inevitable. I think that we should force to use [this](https://github.com/envoyproxy/envoy-wasm/pull/613/files#diff-fefbda479b8fc667205a556beec93b64R1). And warn to use this script by DO NOT EDIT comment by inserting that "We must use tools/cargo-raze.sh".

Shikugawa: or, automatically generate with cargo-raze.sh from pre-push script by checking the Cargo.toml had staged or not.

lizan: Meanwhile, yes, eventually I think it can made into rules_rust to generate WORKSPACE rule, just like rules_python's `pip3_import`.

test/extensions/filters/http/wasm/wasm_filter_test.cc
PiotrSikora: What's the issue? All those examples have the same dependencies, so `cargo raze` should always generate the same `//cargo`. Also, `cargo raze` is not executed as part of the build process, so it shouldn't really influence the location of checked-in `*.rs` and `*.wasm` files.

Having said that, this is me being pedantic, and if upstream doesn't mind and won't complain about this in the upstream PR, then we can merge this. cc @htuch

Shikugawa: In this case, processing artifacts from `cargo raze` is pretty complex. Because these should be placed on envoy specific paths. So I'd like to automate these processes. If we do that when Rust test data are scattered and required special library for new Rust test data in the future, it will be difficult to automate them.

htuch: Is the issue that we're not putting the test `.rs` in the same directory as the test itself? It seems desirable to do this IMHO, otherwise Rust/Wasm is some strange outlier.

htuch: Also, please don't manually construct paths like you do here, if you want runtime files, use https://github.com/envoyproxy/envoy/blob/466245b50e6d475be841e1c8d5f254dd6ea41272/test/test_common/environment.h#L114

Shikugawa: Sure. But with desirable approach, we should manually edit cargo-raze generated scripts. IMHO it might have to be tech debt...., because we can't this process automatically.

test/extensions/bootstrap/wasm/test_data/logging_rust/Cargo.toml
```diff
@@ -10,9 +10,16 @@ proxy-wasm = "0.1"
 log = "0.4"
 
 [lib]
-crate-type = ["cdylib"]
+path = "src/*.rs"
```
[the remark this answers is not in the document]
Shikugawa: Cargo.toml is only to trigger to generate bazel definition. So no need to take care.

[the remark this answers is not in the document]
PiotrSikora: I actually meant that the `path = "src/*.rs"` is not necessary, but my fault for not being clear about this.

bazel/crates.bzl
```diff
@@ -0,0 +1,113 @@
...
+        url = "https://crates-io.s3-us-west-1.amazonaws.com/crates/cfg-if/cfg-if-0.1.10.crate",
+        type = "tar.gz",
+        strip_prefix = "cfg-if-0.1.10",
+        build_file = Label("//bazel/external/cargo/remote:cfg-if-0.1.10.BUILD"),
+    )
+
+    _new_http_archive(
+        name = "raze__hashbrown__0_7_2",
+        url = "https://crates-io.s3-us-west-1.amazonaws.com/crates/hashbrown/hashbrown-0.7.2.crate",
+        type = "tar.gz",
```
htuch, line 46: These are all very similar, can you create a `_rust_crate(..)` macro that factors out all the boilerplate? it should be possible to have just `_rust_crate(name="hashbrown", version="0.7.2")` and it can fill the rest I think.

bazel/external/cargo/remote/ahash-0.3.8.BUILD
htuch: Definitely want a tracking issue for this one, since this is a lot of boilerplate that we should be able to minimize. But I'd be willing to take it as MVP.

test/extensions/filters/http/wasm/test_data/BUILD
```diff
@@ -5,15 +5,68 @@ load(
...
+    deps = [
+        "//bazel/external/cargo:log",
+        "//bazel/external/cargo:proxy_wasm",
+    ],
+)
+
+wasm_rust_binary(
+    name = "shared_queue_rust.wasm",
+    srcs = ["shared_queue_rust/src/lib.rs"],
+    rustc_flags = ["-Clink-arg=-zstack-size=32768"],
```
htuch, line 65: Why do we have this stack size boiler plate?

bazel/crates.bzl
PiotrSikora: Those are all automatically generated by `cargo raze` tool, using macros here would turn this into manual work.

test/extensions/filters/http/wasm/test_data/BUILD
PiotrSikora: There is one test that verifies linear memory size for `test_rust.rs`, but it can be removed from other tests... and `test_rust.rs` can be probably completely removed now that we have tests written using Proxy-Wasm Rust SDK.

Shikugawa: Just deleted unused stack size specification right now. I'm not sure if the `test_rust.rs` is required so missing it.

PR #621, opened 2020-08-16
Fix wasm onLog

gargnupur:
Signed-off-by: gargnupur <gargnupur@google.com>

Commit Message: Don't return in log, if request is not started. We miss cases like localReply in that scenario
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1404,7 +1404,7 @@ void Context::log(const Http::RequestHeaderMap* request_headers,
                   const Http::ResponseTrailerMap* response_trailers,
                   const StreamInfo::StreamInfo& stream_info) {
   if (!http_request_started_) {
-    return;
+    onCreate();
   }
   access_log_request_headers_ = request_headers;
   // ? request_trailers  ?
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1403,9 +1403,14 @@ void Context::log(const Http::RequestHeaderMap* request_headers,
                   const Http::ResponseHeaderMap* response_headers,
                   const Http::ResponseTrailerMap* response_trailers,
                   const StreamInfo::StreamInfo& stream_info) {
-  if (!http_request_started_) {
+  if (!in_vm_context_created_) {
+    // If the request is invalid then onRequestHeaders() will not be called and neither will
+    // onCreate() in cases like sendLocalReply who short-circuits envoy
+    // lifecycle. This is because Envoy does not have a well defined lifetime for the combined HTTP
+    // + AccessLog filter. Thus, to log these scenarios, we call onCreate() in log function below.
     onCreate();
   }
+
   access_log_request_headers_ = request_headers;
   // ? request_trailers  ?
   access_log_response_headers_ = response_headers;


source/extensions/common/wasm/context.cc
```diff
@@ -1404,7 +1404,7 @@ void Context::log(const Http::RequestHeaderMap* request_headers,
                   const Http::ResponseTrailerMap* response_trailers,
                   const StreamInfo::StreamInfo& stream_info) {
   if (!http_request_started_) {
-    return;
+    onCreate();
```
mandarjog, line 1407: I think it is better to have something like 
EnsureContext function, it is very confusing to read that log calls on create.

gargnupur: /cc @jplevyak 

This is because http request is not started...
How about something like this in a "EnsureContext" function?
if (asContext()==nullptr) {
  onCreate();
}

jplevyak: This is a problem with the Envoy interface, so we should not be hacking a patch into the proxy-wasm-cpp-host library.   The contract is that onCreate should be called before any other functions are called which require a VM context are called.  That was an explicit and intentional design decision.  It is up to the Envoy interface code to ensure that all calls into the VM which require a context are bracked with onCreate and onDestroy.  It is confusing to see that log() calls onCreate, but that is because Envoy does not have a well defined lifetime for the combined HTTP + AccessLog filter.   The solution is the work around this limitation by either 1) not calling into the VM if the Context is not already created (this is the solution which we took earlier and which we now have discovered doesn't work for your use case), or 2) create the VM Context on demand when it is needed for the log() case (as is being done here).

This was discussed with Harvey and it was the decision that Wasm should attempt to fixup the lifetime issues for Envoy Filters to the extent possible.  It might be better if we could have a full lifecycle, but sendLocalReply short-circuits that so instead we have a log() call coming without the onRequestHeaders() call which would normally preface it so we can add the onCreate() call here as needed.

jplevyak: Please add a long winded comment about why we are doing this, including an explicit mention of sendLocalReply and the way that it short circuits the lifecycle of the filter.

mandarjog: @jplevyak i understand, in that case it is just naming issue.

If (!context_created) {
  OnCreate();
}
?

mandarjog:
@gargnupur @jplevyak is asan failure real?

source/extensions/common/wasm/context.cc
jplevyak: in_vm_context_created_ is already available in the ContextBase superclass object.   You could use that.

gargnupur: Done... ptal

I changed http_request_started to in_vm_context_created_. @jplevyak : I don't think there are there any cases where http_request_started is false and we don't want to call onLog?

gargnupur:
> @gargnupur @jplevyak is asan failure real?

@mandarjog : looks like not...

jplevyak:
Seems like a timeout @mandarjog

source/extensions/common/wasm/context.cc
kyessenov: Looks fine. Just a couple of questions to make sure it's working as-intended:
- is onDestroy still called in the exceptional case?
- do we need to worry about server-first protocols (onResponse called before onRequest)?

PR #632, opened 2020-08-27
Import envoyproxy and upstream wasm

jplevyak:


jplevyak:
The goal is to keep the upstream PR and envoy-wasm in sync by merging changes up and down as necessary.

jplevyak:
Looks like I need to revert the CI changes.  Are there any we should keep in this repo?

PR #633, opened 2020-09-01
Import envoyproxy and upstream wasm

jplevyak:


jplevyak:
The all_tests failure is a resource issue and lizan says we can't fix it so let's just submit over it.

jplevyak:
All these CI failures seem to be fake and/or coverage which is being worked on.  Can someone verify and I'll merge.

PiotrSikora:
`all_tests` - you can remove or comment out those extra test targets in `ci/do_ci.sh` instead of merging failing tests.

`asan` - you can disable test for integration with external workspaces using `ENVOY_BUILD_FILTER_EXAMPLE` or by removing it in `ci/do_ci.sh` (which I did multiple times, but you keep restoring it in merges).

PiotrSikora:
> `all_tests` - you can remove or comment out those extra test targets in `ci/do_ci.sh` instead of merging failing tests.

Actually, this should be passing, it's either bad merge or build image.

jplevyak:
Everything passes in all_tests, it just doesn't have disk space.  Should I just remove some arbitrary core tests which we should not be effecting?  For the asan build do you have an example of the do_ci.sh change, e.g. SHA that I could back down to?

htuch:
@lizan re: space issues.

lizan:
@jplevyak @htuch I have no good way to fix CircleCI all_test, as azure pipelines is also enabled here, as long as `envoy-wasm (Linux-x64 release) ` is passed we should be good. Can you just remove that Circle job?

kyessenov (approved):

jplevyak:
I tried to kick azure, but they had an outtage which seems to be ongoing.  If anyone knows how to kick this I think it is otherwise good to go.

jplevyak:
clang-tidy seems to be getting stuck.  It normally takes under 10 min and it has been almost 2 hours now.  I  think we have an azure CI issue of some kind.

PR #635, opened 2020-09-02
tests: add more coverage

kyessenov:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/test/extensions/filters/http/wasm/test_data/test_cpp.cc
+++ b/test/extensions/filters/http/wasm/test_data/test_cpp.cc
@@ -128,15 +128,15 @@ bool TestRootContext::onConfigure(size_t) {
     }
     {
       // Some properties are defined in the root context.
-      std::vector<std::pair<std::string, std::string> > properties = {
-        {"plugin_name", "plugin_name"},
-        {"plugin_vm_id", "vm_id"},
-        {"listener_direction", std::string("\x1\0\0\0\0\0\0\0\0", 8)}, // INBOUND
-        {"listener_metadata", ""},
+      std::vector<std::pair<std::vector<std::string>, std::string>> properties = {
+        {{"plugin_name"}, "plugin_name"},
+        {{"plugin_vm_id"}, "vm_id"},
+        {{"listener_direction"}, std::string("\x1\0\0\0\0\0\0\0\0", 8)}, // INBOUND
+        {{"listener_metadata"}, ""},
       };
       for (const auto& property : properties) {
         std::string value;
-        if (!getValue({property.first}, &value)) {
+        if (!getValue(property.first, &value)) {
           logWarn("getValue should provide a value in the root context: " + property.second);
         }
         if (value != property.second) {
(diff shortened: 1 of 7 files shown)


jplevyak:
LGTM but that ASAN error looks interesting.  Seems like the set function is aliasing not copying the data from the VM.

kyessenov:
@jplevyak I think that's because initializer list is somehow destroyed before the call is made? What do you think about adding absl::Span support to SDK?

jplevyak:
We don't have a absl dependency in general now and given that everyone is moving away from absl as the best bits end up in std I would rather try to root cause this.  Abseil is a big dependency for a small bug.

test/extensions/filters/http/wasm/test_data/test_cpp.cc
```diff
@@ -100,6 +101,53 @@ bool TestRootContext::onStart(size_t configuration_size) {
...
+          logWarn("getProperty should not return a value in the root context");
+        }
+      }
+    }
+    {
+      // Some properties are defined in the root context.
+      std::vector<std::pair<std::string, std::string> > properties = {
+        {"plugin_name", "plugin_name"},
+        {"plugin_vm_id", "vm_id"},
+        {"listener_direction", std::string("\x1\0\0\0\0\0\0\0", 8)}, // INBOUND
```
jplevyak, line 134: The problem is right here.  The std::string does not include a null terminator.  The net result is that when getProperty adds the +1 to the span length, it is going to overrun the buffer that std::string has.

jplevyak: I am testing a possible fix : https://github.com/envoyproxy/envoy-wasm/pull/636

kyessenov: Thanks. Pushed a potential fix here. Strangely, bazel test reported "cached" results, meaning the Wasm files are identical with or without the last commit.

kyessenov: Didn't work.

jplevyak: Yah, perhaps you are right, something seems to be odd the initializer_list.

jplevyak:
I think we need to do something more like what is in https://github.com/envoyproxy/envoy-wasm/pull/636

This will require changing the SDK.  I don't think the lifetime of the initializer_list is such that we can count on it outside of an immediate (e.g. foo({...}}  where foo takes an initializer_list because it is just a set of aliases to potentially ephemeral things.

PTAL and tell me what you think.

kyessenov:
@jplevyak I'm not a fan of using vectors just to build a string. I was thinking of the following:
1. Pre-define a bunch of {...} properties as constant, and it would be getValue(string, &value). This approach is a bit high maintenance since properties would need to be enumerated.

2. Accept an optional absl::Span version of this interface so both initializer lists and vectors can be passed.

WDYT?

jplevyak:
The problem isn't the lack of Span.  The problem is that initailizer_list is only an alias and you can't define one over another object which isn't going to be alive when it is used.  Span is not going to fix that problem.  Either it will alias and we have another ASAN error or it will materialize and we will have the equivalent of std::vector.

You just can't define an initializer_list as a variable pointing to some ephemeral local and then use it later.

This is the same as 

Foo& foo = *(std::make_unique<Foo>());
... use foo.

It just doesn't make sense.

jplevyak:
OK.  Turn out we can mix the initializer_list and std::vector versions and function overloading will pick the right ones.  The trick is that the initializer_list must be const&.  See  https://github.com/envoyproxy/envoy-wasm/pull/636 for details.

If you like this version I can add the new versions of getValue and getProperty to the SDK.

PTAL

kyessenov:
Yeah, it looks good. Just want to make sure there is no vector being constructed most of the time.

jplevyak:
According to https://stackoverflow.com/questions/42613660/how-does-the-compiler-decide-between-vector-vs-initializer-list this solution should pick the correct ones, and if I remove the const& from the initializer_list it fails to compile so I am confident that it isn't finding the std::vector versions instead of the initializer_list versions in the case of non-vector immediates.

jplevyak (approved):

PR #644, opened 2020-09-23
Make wasm/ext targets visible,

jplevyak:
Make wasm/ext targets visible,


kyessenov (approved):

kyessenov:
Format error is something else.

PR #649, opened 2020-09-29
Fix jslib target for wasm binaries.

jplevyak:

### DIFF
--- a/test/extensions/access_loggers/wasm/test_data/BUILD
+++ b/test/extensions/access_loggers/wasm/test_data/BUILD
@@ -3,7 +3,7 @@ load(
     "envoy_cc_library",
     "envoy_package",
 )
-load("//bazel/wasm:wasm.bzl", "wasm_cc_binary")
+load("//bazel/wasm:wasm.bzl", "envoy_wasm_cc_binary")
 
 licenses(["notice"])  # Apache 2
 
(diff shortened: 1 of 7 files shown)


PiotrSikora (approved):

test/extensions/access_loggers/wasm/test_data/BUILD
```diff
@@ -26,7 +26,7 @@ envoy_cc_library(
     ],
 )
 
-wasm_cc_binary(
+envoy_wasm_cc_binary(
```
PiotrSikora, line 29: This could have stayed `wasm_cc_binary` with `repository = "@envoy"`, instead of creating a new rule, IMHO.

PiotrSikora (approved):

test/extensions/access_loggers/wasm/test_data/BUILD
jplevyak: Then I would have to add that to every target.   This is simpler IMO.

PiotrSikora: Correct, and I agree, but we do the same (i.e. add `repository = "@envoy"` to every target) for `envoy_cc_library`, etc., so this way it would be at least consistent.

jplevyak: That is the great thing about standards, there are so many to choose to be consistent with.  I choose the envoy_cc_library standard which adds custom envoy logic into a wrapper function rather than the repository="@envoy" standard.  I think the latter is a bug and should be fixed :)

PiotrSikora: So you're choosing to be consistently inconsistent? :)

(I approved this PR when making the initial comment, so feel free to merge it as-is, I don't really care)

PR #655, opened 2020-10-08
fix crash on failed wasm

kyessenov:

Fixes https://github.com/istio/istio/issues/27685


PiotrSikora:
For istio/istio#27685, shouldn't the correct fix be to not load the extension at all (i.e. envoyproxy/envoy-wasm/issues/252)?

kyessenov:
@PiotrSikora AFAICT it's not loaded. Envoy requires response details any team stream is reset.

kyessenov:
@PiotrSikora It would be nice if the filter would reject the config. But it's not happening, so not crashing is better than crashing.

PiotrSikora (approved):

PR #656, opened 2020-10-08
Fix tests in previous commit.

PiotrSikora:


kyessenov (approved):

kyessenov:
I think the test is not cleaning up the stream. The assertion should be triggered.

PiotrSikora:
Which assertion?

kyessenov:
There's an assertion that responseCodeDetails are set.

PiotrSikora:
Right, but in the previous commit you've added code to set responseCodeDetails if it's missing, so it's set now, isn't it?

kyessenov:
Yes, it is set but it's not tested. I am just suggesting to fix the test and make it clean up explicitly, so that assertion is exercised. We wouldn't have the issue to begin with if the test was right.

PiotrSikora:
I thought your code fixed issue in production and not tests?

PiotrSikora:
I'm going to merge this to fix master, we can discuss proper fix later.

